PR #5, opened 2025-06-29
more intuitive layout, plus an example to explain the intuition

HanGuo97:
No functionality change (as we swap both shape and stride), but rather improve the readability of layout

### DIFF
--- a/quack/reduction_base.py
+++ b/quack/reduction_base.py
@@ -39,17 +39,32 @@ class ReductionBase:
         vecsize = copy_bits // self.dtype.width
         assert self.N % vecsize == 0, f"Input N {self.N} is not divisible by vector size {vecsize}"
         num_threads = self._get_num_threads()
-        num_warps = num_threads // cute.arch.WARP_SIZE
         assert num_threads % cute.arch.WARP_SIZE == 0
 
         threads_per_row = self._calculate_threads_per_row()
         num_blocks_N = cute.ceil_div(self.N // vecsize, threads_per_row * self.cluster_n)
         cols_per_block = num_threads // threads_per_row
+
+        # Example with `N = 512` and FP16, then each tile has `[4, 512]` elements.
+        # - `vector_size = 8` (i.e., 16-bit x 8 = 128-bit)
+        # - `num_threads = 128`
+        # - `threads_per_row = 32`
+        # - `cols_per_block = 4`
+        # - `cluster_n = 1`
+        # - `num_blocks_N = 2`
+        # TV Layout:
+        # - Shape  (4, 32), (8, 2)
+        # - Stride (1, 32), (4, 1024)
+        # Here's the memory layout:
+        # - Row 0: [B0, T0 x8] [B0, T4 x8] ... [B0, T124 x8] [B1, T0 x8] [B1, T4 x8] ... [B1, T124 x8]
+        # - Row 1: [B0, T1 x8] [B0, T5 x8] ... [B0, T125 x8] [B1, T1 x8] [B1, T5 x8] ... [B1, T125 x8]
+        # - Row 2: [B0, T2 x8] [B0, T6 x8] ... [B0, T126 x8] [B1, T2 x8] [B1, T6 x8] ... [B1, T126 x8]
+        # - Row 3: [B0, T3 x8] [B0, T7 x8] ... [B0, T127 x8] [B1, T3 x8] [B1, T7 x8] ... [B1, T127 x8]
         tiler_mn = (cols_per_block, vecsize * num_blocks_N * threads_per_row)
         tv_layout = cute.make_layout(
-            ((threads_per_row, cols_per_block), (vecsize, num_blocks_N)),
+            ((cols_per_block, threads_per_row), (vecsize, num_blocks_N)),
             stride=(
-                (vecsize * cols_per_block, 1),
+                (1, vecsize * cols_per_block),
                 (cols_per_block, cols_per_block * vecsize * threads_per_row),
             ),
         )


quack/reduction_base.py
```diff
@@ -39,17 +39,32 @@ def _get_tv_layout(self):
...
+        # - Stride (1, 32), (4, 1024)
+        # Here's the memory layout:
+        # - Row 0: [B0, T0 x8] [B0, T4 x8] ... [B0, T124 x8] [B1, T0 x8] [B1, T4 x8] ... [B1, T124 x8]
+        # - Row 1: [B0, T1 x8] [B0, T5 x8] ... [B0, T125 x8] [B1, T1 x8] [B1, T5 x8] ... [B1, T125 x8]
+        # - Row 2: [B0, T2 x8] [B0, T6 x8] ... [B0, T126 x8] [B1, T2 x8] [B1, T6 x8] ... [B1, T126 x8]
+        # - Row 3: [B0, T3 x8] [B0, T7 x8] ... [B0, T127 x8] [B1, T3 x8] [B1, T7 x8] ... [B1, T127 x8]
         tiler_mn = (cols_per_block, vecsize * num_blocks_N * threads_per_row)
         tv_layout = cute.make_layout(
-            ((threads_per_row, cols_per_block), (vecsize, num_blocks_N)),
+            ((cols_per_block, threads_per_row), (vecsize, num_blocks_N)),
```
tridao, line 65: After thinking about it more I realize the new tv_layout is not equivalent to the old tv_layout (and running the kernels with this new layout will produce different results). Note that this is tv_layout and not a regular layout, so the "thread" part need to correspond to the "value" part.

In your example, each thread to read (8, 2) values, like this:
Row 0: [T0 x 8] [T1 x 8] ... [T31 x 8] [T0 x 8] [T1 x 8] ... [T31 x 8]
Row 1: [T32 x8] [T31 x8] ...
Row 2: [T64 x 8] [T65 x 8] ...
Row 3: [T96 x 8] [T97 x 8] ...

To back out the tv_layout: 
- We want T0 -> T1 to move by vecsize=8 units in the N dimension. The N dimension has stride `cols_per_block`=4. So the first stride of the "thread" part of the tv_layout is cols_per_block * 8.
- We want T0 -> T32 to moe by 1 unit in the M dimension. M dimension has stride 1. So the 2nd stride of the "thread" part of the tv_layout is 1.
- Similarly we can work out the strides of the "value" part to be (cols_per_block, cols_per_block * vecsize * threads_per_row)

jeromeku:
The thread-value ownership changes with the new layout while the overall tile shape remains the same.

Not sure how this impacts downstream operations depending on how you re-partition the destination tensor.

Here is the previous TV layout:
[old_tile.pdf](https://github.com/user-attachments/files/21068735/old_tile.pdf)

The new layout:
[new_tile.pdf](https://github.com/user-attachments/files/21068737/new_tile.pdf)

The main difference is that in the previous layout, the thread ids are `row-major` within block while in the new layout the tids are `column-major` within block.

```cpp
#include "cute/tensor.hpp"
using namespace cute;

constexpr int N = 512; // Change to 256 if latex can't handle

using Element = cute::half_t;
constexpr int ACCESS_SIZE_BITS = 128;
constexpr int VEC_SIZE = ACCESS_SIZE_BITS / cute::sizeof_bits_v<Element>;

constexpr int NUM_THREADS = 128;
constexpr int THREADS_PER_ROW = 32;
constexpr int COLS_PER_BLOCK = NUM_THREADS / THREADS_PER_ROW;
constexpr int CLUSTER_N = 1;
constexpr int NUM_BLOCKS_N = cute::ceil_div(N / VEC_SIZE, THREADS_PER_ROW * CLUSTER_N);

using Thr_Shape_new = Shape<Int<COLS_PER_BLOCK>, Int<THREADS_PER_ROW>>;
using Thr_Shape_old = Shape<Int<THREADS_PER_ROW>, Int<COLS_PER_BLOCK>>;
using Val_Shape = Shape<Int<VEC_SIZE>, Int<NUM_BLOCKS_N>>;

using Thr_Stride_new = Stride<Int<1>, Int<VEC_SIZE * COLS_PER_BLOCK>>;
using Thr_Stride_old = Stride<Int<VEC_SIZE * COLS_PER_BLOCK>, Int<1>>;
using Val_Stride = Stride<Int<COLS_PER_BLOCK>, Int<COLS_PER_BLOCK * VEC_SIZE * THREADS_PER_ROW>>;

using TV_Shape_old = Shape<Thr_Shape_old, Val_Shape>;
using TV_Stride_old = Stride<Thr_Stride_old, Val_Stride>;
using TV_Layout_old = Layout<TV_Shape_old, TV_Stride_old>;

using TV_Shape_new = Shape<Thr_Shape_new, Val_Shape>;
using TV_Stride_new = Stride<Thr_Stride_new, Val_Stride>;
using TV_Layout_new = Layout<TV_Shape_new, TV_Stride_new>;

using TV_Layout = TV_Layout_old;
using MN_Tiler = Shape<Int<COLS_PER_BLOCK>, Int<VEC_SIZE * NUM_BLOCKS_N * THREADS_PER_ROW>>;

int main()
{
    // Thread arrangement
    using Element = half_t;

    using CopyOp = UniversalCopy<uint_byte_t<VEC_SIZE>>; // A very specific access width copy instruction
    using Atom = Copy_Atom<CopyOp, Element>;

    TiledCopy tiled_copy = make_tiled_copy_impl(Atom{}, TV_Layout{}, MN_Tiler{});
    print_latex(tiled_copy);

}
```

HanGuo97:
Closing it as I don't know how to improve the readability otherwise.

PR #10, opened 2025-07-14
Layernorm Backwards

simveit:
Hello,

In this PR I am trying to implement Layernorm backward pass. To keep things simple I first tried to implement without the double buffer that is used in RMSNorm backwards.

The following tests pass:
```
@pytest.mark.parametrize("N", [256, 2048, 32768, 65536])])
```
However when I try non potences of 2 like the below it fails
```
@pytest.mark.parametrize("N", [760, 1128])  # , 32768])
```

could someone explain to me why that is?


simveit:
Recent version fixed the problem above.
This needs works now, still need to adopt benchmark script and potentially smem size therefore stays a draft for now.

PR #11, opened 2025-07-15
clearer _calculate_threads_per_row and _set_cluster_n  (if-elif-else rather than nested ternaries)

lessw2020:
Fantastic kernels here! 
This PR:
1 - improves readability and hopefully maintainability of 
~~~
_calculate_threads_per_row 
~~~
and 
~~~
_set_cluster_n 
~~~
by moving them to if-elif-else blocks rather than the current nested ternaries.

I found the nested ternaries hard to follow/read, and figure others may find the same.  Hopefully this also improves maintainability.

Testing:
Ran Pytest tests/test_rmsnorm, 
exact same results as original (337 pass, 3 fail, 2 skipped)

### DIFF
--- a/quack/rmsnorm.py
+++ b/quack/rmsnorm.py
@@ -1,14 +1,14 @@
 # Copyright (c) 2025, Wentao Guo, Ted Zadouri, Tri Dao.
 
-import torch
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


tridao:
Thanks!

PR #15, opened 2025-07-15
Add reload to rmsnorm backward, fix tests

simveit:
1) Some of the tests where broken, I fixed that

2) Added Reload in rmsnorm backward which brings small but measurable benefit for some cases:

`M = 32768`:

With reload from `self.reload_wdy`:
8192: `2834.00`
16384: `2566.00`
32768: `2470.00`
65536: `2407.00`
131072: `2145.00`

Without reload from `self.reload_wdy`:
8192: `2833.00`
16384: `2578.00`
32768: `2454.00`
65536: `2340.00`
131072: `2147.00`

Sadly reloading of `x_hat` doesn't bring any benefit, I tried that out aswell.


tridao:
thank you!

Issue #12, opened 2025-07-15
Integrating Quack.RMSNorm into real training codebase - fp32 for weights is not realistic (only for reference testing?), needs to accept 3D inputs


lessw2020:
I'm currently integrating the quack.rmsnorm kernel into TorchTitan to showcase these impressive kernels.

However, during integration I hit two key issues;
a -  RMSNorm has a hard assert that the weights are fp32?  This seems unrealistic for real world use...bf16 works nicely and is the default for most. 
b - RMSNorm also has a hard assert on 2D inputs...but most inputs for actual training are 3D.  This one is not too hard to resolve but I think it would be good to integrate this into the kernel wrapper to resolve.

For fp32 - I wanted to check if there is some other reason I'm not aware of why fp32 is enforced (assuming it may just be b/c much easer to verify results vs reference, using fp32 for testing, but now that it's moving to real world use can we lift this?)

tridao:
No real reason why weight should be in fp32, I usually use torch AMP for small models where rmsnorm weights are fp32. Should be easy to change. The only place that needs attention is that we still want to compute the partial weight gradient in fp32 before summing up, then convert that to bf16. I might take a look a bit later.
3D input should also be easy to change, just a reshape at the beginning and end of the function (in pytorch).

lessw2020:
Thanks @tridao  - I've made a PR to handle the 3D input portion.
https://github.com/Dao-AILab/quack/pull/13

Would be great if you can update the kernel for bf16 support.   Otherwise, I'll take a crack at it tomorrow paying attention to your comments above.

lessw2020:
I've made a PR (https://github.com/Dao-AILab/quack/pull/16) that updates support for weights in BF16.  
Let me go ahead and also add in support for fp16 as well for completeness!

lessw2020:
We are now up and running in TorchTian via draft pr, and both issues above have been resolved via landed quack prs...thus closing this tracking issue.

PR #16, opened 2025-07-15
Add RMSNorm support for BF16 and FP16 weights, add unit tests to verify BF16, FP16 weight results for fwd, bwd

lessw2020:
Based on issue (https://github.com/Dao-AILab/quack/issues/12), this PR adds:
1 - Support for RMSNorm weights in BF16, FP16, or FP32.  Previous code was limited to weights only in Fp32. 
Note that per @tridao, I've implemented it such that partial weight gradients are always computed in fp32. 
("we still want to compute the partial weight gradient in fp32 before summing up, then convert that to bf16")

* Now running in TorchTitan with FSDP2 and successfully training! 
* Speedup on LLama3-8B, B200*8, FSDP2 with TorchTitan:
Screenshot 2025-07-15 at 4 06 37 PM

2 - adds unit tests to verify BF16 and FP16 weights produce matching results to reference, for both forward and backwards. 
*   Forward pass tests:
    *   test_rmsnorm_bf16_weights 
    *   test_rmsnorm_fp16_weights 
*   Backward pass tests:
    *   test_rmsnorm_bf16_weights_backward 
    *   test_rmsnorm_fp16_weights_backward. 
    
 2a - update input validation test to verify fp16, bf16 and fp32 are accepted.  
 use fp64 input to test bad weight input. 
    
3 - compile cache is also updated to key on weight dtype as well. 

Testing:
a - verified all previous unit tests passing as before (337 passing)
b - add and run new unit tests (test_rmsnorm_bf16_weights is passing.
Screenshot 2025-07-15 at 11 05 18 AM
Screenshot 2025-07-15 at 12 22 42 PM

In TorchTitan:

#### =====================================================
Median Tokens/Second (excluding step 1): 9854.0
Max Memory Usage: 80.17 GiB

####  =====================================================
Median Tokens/Second (excluding step 1): 8752.0
Max Memory Usage: 90.37 GiB

### DIFF
--- a/quack/rmsnorm.py
+++ b/quack/rmsnorm.py
@@ -257,7 +257,8 @@ def _rmsnorm_fwd(
     assert weight.dtype in [
         torch.float32,
         torch.bfloat16,
-    ], "Weight must be float32 or bfloat16"
+        torch.float16,
+    ], "Weight must be float32, float16 or bfloat16"
 
     M, N = x.shape
     device = x.device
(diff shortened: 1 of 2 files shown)


quack/rmsnorm.py
```diff
@@ -773,3 +809,39 @@ def rmsnorm(x: torch.Tensor, weight: torch.Tensor, eps: float = 1e-6) -> torch.T
         Normalized output tensor of same shape as x
     """
     return RMSNormFunction.apply(x, weight, eps)
+
+
+class QuackRMSNorm(torch.nn.Module):
```
tridao, line 814: we can just subclass torch.nn.RMSNorm instead? and only change the forward?

quack/rmsnorm.py
```diff
@@ -632,9 +642,22 @@ def kernel(
                     tXsdW_other = cute.make_tensor(tXsdW.iterator + i * sdW.stride[0], tXsdW.layout)
                     cute.autovec_copy(tXsdW_other, tXrdW_other)
                     tXrdW.store(tXrdW.load() + tXrdW_other.load())
-                cute.copy(copy_atom_store_dW, tdWrdW, tdWgdW, pred=tdWpdW)
+                # Only convert for storing to weight.dtype if weights are not fp32
```
tridao, line 645: `tdWgdW.element_type` should always be fp32, since we allocate dw_partial to be in fp32. 
You can just put an assert here that type is fp32.

tridao:
Wow amazing that you're already seeing speedup in torch titan!

quack/rmsnorm.py
lessw2020: in theory yes, but I started trying this tonight and hit various cute dsl issues in the process.
Thus, I'd like to push this off as a seperate PR rather than try and do it here as we already have a lot of changes with bf16/fp16 support coming in. 
subclassing may also give us free torch.compile support as that's a different problem atm (we have to provide a dummy wrapper for compile to not try and dive into the quack code). 
Opened Issue https://github.com/Dao-AILab/quack/issues/18 to track torch.compile issue (and by proxy, rmsnorm subclassing)

tridao (approved):

PR #20, opened 2025-07-16
add reduction options to Quack Cross Entropy for PyTorch drop-in usage

lessw2020:
This PR make Quack Cross Entropy loss 'drop in ready' for PyTorch models.  It does this by adding:
1 - options to reduce the cross entropy loss based on mean, sum, or none. 

Currently the cross entropy returns a tensor of per example losses, but if you try and drop it into a PyTorch model and go, you will encounter:
~~~
RuntimeError: grad can be implicitly created only for scalar outputs
~~~

This is because most people are calling loss.backward() expecting a scalar return value in their training loop. 

This PR resolves that with the caveat that you still have to spec 'reduction = 'mean' when instantiating Quack CE within PyTorch, but now it's much easier to do.   
Not sure if I should add a note in the comments about usage...

#### Perf:

With that change, I tested on B200 and Llama3-8B, eager:
Screenshot 2025-07-16 at 3 45 18 PM

Not much speedup, but a nice drop in Memory.  Note that combinining this with Quack RMSNorm and you get a meaningful **26.29%** drop in memory and speedup of **~12%**.  Thus, some synergy that is meaningful esp for memory.

#### Testing:
The defaults are unchanged, and as expected, all cross entropy testing passes:

Screenshot 2025-07-16 at 3 38 43 PM

### DIFF
--- a/quack/cross_entropy.py
+++ b/quack/cross_entropy.py
@@ -1,16 +1,16 @@
 # Copyright (c) 2025, Wentao Guo, Ted Zadouri, Tri Dao.
 
 import math
-import torch
 from typing import Optional, Type
 
 import cuda.bindings.driver as cuda
 
 import cutlass
 import cutlass.cute as cute
-from cutlass.cute.runtime import from_dlpack
 
 import quack.utils as utils
+import torch
+from cutlass.cute.runtime import from_dlpack
 from quack.reduction_base import ReductionBase, torch2cute_dtype_map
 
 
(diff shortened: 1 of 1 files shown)


lessw2020:
Note - pondering if an enum is better for the reduction option (Rust side in me says yes, but this is Python) so appreciate feedback on leave as string or add enum.

tridao:
Let's keep it as string to match pytorch

PR #27, opened 2025-08-07
Symmetric alpha * A @ B + beta * C

JackCharlesZhang:
Implement dense gemm for output matrices that are guaranteed to be square and symmetric.

### DIFF
--- a/quack/symmetric_dense_gemm_sm90.py
+++ b/quack/symmetric_dense_gemm_sm90.py
@@ -1728,16 +1728,14 @@ def run(
             ),
         ).to(torch_dtype)
 
-        # Create symmetric matrix: A @ A^T, one layer at a time to prevent OOM
-        symmetric_matrices = []
-        for batch_idx in range(l):
-            matrix_slice = base_tensor[batch_idx] 
-            # Create symmetric matrix: A @ A^T (shape: mode0 x mode0)
-            symmetric_matrix = torch.matmul(matrix_slice, matrix_slice.transpose(-2, -1))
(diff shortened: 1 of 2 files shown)


quack/symmetric_dense_gemm_sm90.py
```diff
@@ -0,0 +1,2035 @@
...
+    timing = do_bench(fn, warmup=warmup, rep=repeats)
+    tflops = flops / (timing * 1e9)  # Convert to TFlops
+    print(f"Cute-DSL Average time: {timing:.3f} ms, TFLOPS: {tflops:.1f}")
+
+    time.sleep(0.5)
+    timing_cublas = do_bench(fn_cublas, warmup=warmup, rep=repeats)
+    tflops_cublas = flops / (timing_cublas * 1e9)  # Convert to TFlops
+    print(f"CuBLAS Average time: {timing_cublas:.3f} ms, TFLOPS: {tflops_cublas:.1f}")
+
+def _symmetric_dense_gemm(
```
tridao, line 1890: Let's support passing in both A and B.
There are cases we care about where A @ B is symmetric but a != b. One example is if A is symmetric and B = A @ A. Then A @ B will be symmetric (you can check)

tridao:
From a quick pass it's looking pretty good to me!
One thing we can do (not in this PR but maybe the next) is to subclass the HopperWgmmaGemm instead of copy-pasting.

quack/symmetric_dense_gemm_sm90.py
JackCharlesZhang: yes, good call

PR #38, opened 2025-09-09
RMS with bias + tests

JackCharlesZhang:
Adding bias to RMS Mean + Tests for bias, residual, amp bf16 training + fixed cache


quack/rmsnorm.py
```diff
@@ -95,6 +94,7 @@ def __call__(
         self,
         mX: cute.Tensor,
         mW: cute.Tensor,
+        mB: cute.Tensor,
```
tridao, line 97: Optional

PR #39, opened 2025-09-10
[WIP] Add Prenorm to RMSNorm

JackCharlesZhang:
Last normalization in the transformer block doesn't have an output residual.

### DIFF
--- a/quack/rmsnorm.py
+++ b/quack/rmsnorm.py
@@ -1105,7 +1105,7 @@ def rmsnorm_bwd(
 
 class RMSNormFunction(torch.autograd.Function):
     @staticmethod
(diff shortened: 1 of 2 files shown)


quack/rmsnorm.py
```diff
@@ -1127,7 +1127,8 @@ def forward(ctx, x, weight, bias=None, residual=None, out_dtype=None, residual_d
         ctx.eps = eps
         ctx.x_shape_og = x_shape_og
         ctx.residual_dtype = residual.dtype if residual is not None else None
-        if residual is None:
+        ctx.prenorm = prenorm
+        if residual_out is None or prenorm == False:
```
JackCharlesZhang, line 1131: use residual_out is None to account for this case in the forward

if residual_out is None:
        residual_out = x
    return out, residual_out, rstd

PR #42, opened 2025-09-16
[GemmSm90] Add cpu varlen m and k

zfan2356:
Hello~ To meet some communication library interface such as `DeepEP`, I'd like to share an alternative approach for passing sequence lengths (`cu_seqlens_m`, `cu_seqlens_k`) to `GemmSm90` using `List[cutlass.Int32]` that can directly access tensor data from CPU memory.

I've conducted some benchmark tests, and the results show that this method achieves comparable throughput to using GPU-resident `cu_seqlens` under the same condition(tile_shape and cluster shape), demonstrating its effectiveness while providing more flexibility in memory management.

```
~/quack$ python benchmarks/benchmark_gemm_sm90.py --varlen_m --mnkl=2048,2048,2048,10 --tile_shape_mnk=128,128,64 --persistent --pingpong --use_cpu_varlen_m
Running Hopper Dense GEMM with:
mnkl: (2048, 2048, 2048, 10)
A dtype: BFloat16, B dtype: BFloat16, D dtype: BFloat16, C_dtype: None, Acc dtype: Float32
Matrix majors - A: k, B: k, D: n
Tile Shape: (128, 128, 64), Cluster Shape: (1, 1)
Tolerance: 0.03
Warmup iterations: 5
Iterations: 30
Skip reference checking: False
Cute-DSL Average time: 0.347 ms, TFLOPS: 494.7, GB/s: 725
PASS
~/quack$ python benchmarks/benchmark_gemm_sm90.py --varlen_k --mnkl=2048,2048,2048,10 --tile_shape_mnk=128,128,64 --persistent --pingpong --use_cpu_varlen_k
Running Hopper Dense GEMM with:
mnkl: (2048, 2048, 2048, 10)
A dtype: BFloat16, B dtype: BFloat16, D dtype: BFloat16, C_dtype: None, Acc dtype: Float32
Matrix majors - A: k, B: k, D: n
Tile Shape: (128, 128, 64), Cluster Shape: (1, 1)
Tolerance: 0.03
Warmup iterations: 5
Iterations: 30
Skip reference checking: False
Cute-DSL Average time: 0.350 ms, TFLOPS: 490.2, GB/s: 718
PASS
```

### DIFF
--- a/benchmarks/benchmark_gemm_sm90.py
+++ b/benchmarks/benchmark_gemm_sm90.py
@@ -341,14 +341,12 @@ def run(
         cu_seqlens_m = torch.arange(0, l + 1, dtype=torch.int32, device="cuda") * m
         mCuSeqlensM = from_dlpack(cu_seqlens_m, assumed_align=4).mark_layout_dynamic(leading_dim=0)
         if use_cpu_varlen_m:
-            mCuSeqlensMCpu = list(range(0, l + 1)) * m
-        else:
-            mCuSeqlensMCpu = None
+            mCuSeqlensM = list(range(0, (l + 1) * m, m))
         if gather_A:
             a_idx_reshaped = rearrange(a_idx_reshaped, "m l -> (l m)")
             mAIdx = from_dlpack(a_idx_reshaped, assumed_align=4).mark_layout_dynamic(leading_dim=0)
     else:
-        cu_seqlens_m, mCuSeqlensM, mCuSeqlensMCpu = None, None, None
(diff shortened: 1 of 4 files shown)


zfan2356:
> Could we store `mCuSeqlensMCpu` in the same tensor as `mCuSeqlensM`? Then we can use if statement based on the type to extract the fragment from `mCuSeqlensM` if necessary?

Of course! I'll fix it soon. Moreover, there seem to be some issues here, and I will verify them with the Cutlass community.

PR #46, opened 2025-09-30
Refactor Symmetric GEMM to be an activation

JackCharlesZhang:
Maintain 2x speedup

Screenshot 2025-09-30 at 6 10 14 PM
Screenshot 2025-09-30 at 6 21 04 PM

### DIFF
--- a/quack/gemm_interface.py
+++ b/quack/gemm_interface.py
@@ -904,9 +904,9 @@ def gemm_dgated_ref(
 )
 def gemm_symmetric_out(
     A: Tensor,  # (M, K) or (L, M, K)
-    B: Tensor,  # (K, N) or (L, K, N)
-    out: Tensor,  # (M, N) or (L, M, N)
-    C: Optional[Tensor] = None,  # (M, N) or (L, M, N)
+    B: Tensor,  # (K, M) or (L, K, M)
+    out: Tensor,  # (M, M) or (L, M, M)
+    C: Optional[Tensor] = None,  # (M, M) or (L, M, M)
     dynamic_scheduler: bool = False,
     alpha: float = 1.0,
     beta: float = 1.0,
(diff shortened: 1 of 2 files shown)


quack/gemm_interface.py
```diff
@@ -895,6 +896,94 @@ def gemm_dgated_ref(
...
+@torch.library.custom_op(
+    "quack::gemm_symmetric_out",
+    mutates_args=("preact_out", "postact_out"),
+    device_types="cuda",
+    schema="(Tensor A, Tensor B, Tensor(a2!)? preact_out, Tensor(a3!) postact_out, Tensor? C=None, bool dynamic_scheduler=False, float alpha=1.0, float beta=1.0) -> ()",
+)
+def gemm_symmetric_out(
+    A: Tensor,  # (M, K) or (L, M, K)
+    B: Tensor,  # (K, N) or (L, K, N)
+    preact_out: Tensor,  # (M, N) or (L, M, N)
```
tridao, line 908: there should just be "out", no "preact_out" or "postact_out". From a user perspective, it shouldn't matter how it's implemented

quack/symmetric_gemm.py
```diff
@@ -0,0 +1,157 @@
...
+class GemmSymmetricSm90(GemmSymmetricMixin, GemmSm90):
+    pass
+
+class GemmSymmetricSm100(GemmSymmetricMixin, GemmSm100):
+    pass
+
+def gemm_symmetric(
+    A: Tensor,  # (l, m, k) or (total_m, k) if varlen_m or (whatever, k) if gather_A with varlen_m
+    B: Tensor,  # (l, n, k)
+    D: Optional[Tensor],  # (l, m, n) or (total_m, n) if varlen_m
```
tridao, line 29: This should take in "D" but no "PostAct". Then internally inside this function we construct "PostAct" as the trnapose of "D".

tridao:
LGTM!

quack/symmetric_gemm.py
```diff
@@ -0,0 +1,149 @@
...
+class GemmSymmetricSm100(GemmSymmetricMixin, GemmSm100):
+    pass
+
+def gemm_symmetric(
+    A: Tensor,  # (l, m, k)
+    B: Tensor,  # (l, m, k)
+    D: Optional[Tensor],  # (l, m, m)
+    C: Optional[Tensor],  # (l, m, m) 
+    tile_count_semaphore: Optional[Tensor],  # (1,)
+    activation: Optional[str],
```
tridao, line 32: I forgot, `activation` shouldn't be an argument here, it should be set to None inside the this function. Let's do it in a followup PR

PR #64, opened 2026-01-08
fix bug in online_softmax when all loaded values in a warp are -inf

LitLeo:
if max_x=-inf, exp_x=nan, which causes error result.

```python
@cute.jit
def online_softmax_reduce(
    x: cute.TensorSSA,
    threads_per_row: cutlass.Constexpr[int],
    reduction_buffer: Optional[cute.Tensor] = None,
    mbar_ptr: Optional[cute.Pointer] = None,
    hook_fn: Optional[Callable] = None,
    phase: Optional[Int32] = None,
    return_exp_x: bool = False,
) -> [Float32, Float32, Optional[cute.TensorSSA]]:
    assert x.dtype == Float32, "x must be of type Float32"
    """reduction_buffer must have shape (num_warps / warps_per_row, (warps_per_row, cluster_n), 2)"""
    max_x = cute.arch.warp_reduction(
        x.reduce(cute.ReductionOp.MAX, init_val=-Float32.inf, reduction_profile=0),
        cute.arch.fmax,
        threads_in_group=min(threads_per_row, cute.arch.WARP_SIZE),
    )
    log2_e = math.log2(math.e)
    exp_x = cute.math.exp2(x * log2_e - (max_x * log2_e), fastmath=True)
    sum_exp_x = cute.arch.warp_reduction(
        exp_x.reduce(cute.ReductionOp.ADD, init_val=0.0, reduction_profile=0),
        operator.add,
        threads_in_group=min(threads_per_row, cute.arch.WARP_SIZE),
    )

### DIFF
--- a/quack/reduce.py
+++ b/quack/reduce.py
@@ -140,13 +140,11 @@ def online_softmax_reduce(
         cute.arch.fmax,
         threads_in_group=min(threads_per_row, cute.arch.WARP_SIZE),
     )
+    if max_x == -Float32.inf:
+        max_x = Float32.zero
 
     log2_e = math.log2(math.e)
     exp_x = cute.math.exp2(x * log2_e - (max_x * log2_e), fastmath=True)
-    # if max_x=-inf, exp_x=nan, which causes error result.
-    if max_x == -Float32.inf:
-        exp_x = cute.zeros_like(x)  
-        
     sum_exp_x = cute.arch.warp_reduction(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by tridao** [comment]
I think it's better to have the check:
```
max_x_cur = Float32(0.0) if max_x == -Float32.inf else max_x
```
The subtracting `max_x_cur` before the exp.

tridao:
I prefer changing max-cur. For the case where the input is entirely -inf (not out of bounds) we still want output to be zero.

PR #73, opened 2026-02-05
[Packaging / docs] Make `pip` install use the torch cu130

vincentzed:
thttps://github.com/Dao-AILab/quack/commit/0934868478610ab06c7d1695004be9e7dcde9adf

Here introduce Cu13, but to enable proper pip install, the editable install shall specify: torch cu130 in the pyproject toml

Also seems whl is not has cu13? (To my understanding)- can we have cu130 whl? It's fast to install, but it would be convenient from whl in the gh releases as well.


tridao (approved):

tridao:
Thanks!

PR #75, opened 2026-02-17
Better gemm configs that reduce 80% autotuning time

GarlGuo:
I benchmarked 18 different MoE configs on H100 and B300 GPU and checked all autotuning logs in quack. The new configs can match the original throughput while reducing 80% autotuning time.


tridao:
LGTM, can merge when it's ready

PR #80, opened 2026-03-11
CLC should be autotunable in SM100 rather than a fixed argument

GarlGuo:
CLC scheduler wins over static tile scheduler for 95% configs, but there are still cases where CLC tile scheduler have worse perf than static tile scheduler.


tridao:
@GarlGuo let's fix this and get this in? We're cutting a release for quack soon

tridao (approved):

PR #99, opened 2026-04-07
add _get_mma_inst_tile_k() virtual method to GemmSm100

santoshmo:
Problem: Subclassing GemmSm100 for specialized GEMM variants (e.g., MoE grouped GEMM) requires overriding the MMA K-tile depth. Currently mma_inst_tile_k = 4 is hardcoded inside _setup_attributes, which feeds into downstream SMEM layout, pipeline stage, and cluster shape calculations. 

In the MoE kernel, each mainloop iteration has additional overhead that doesn't exist in the standard GEMM.

There's no way to change it without either:
- Copying the entire _setup_attributes (~150 lines) into the subclass
- Monkey-patching after _setup_attributes, which breaks because downstream calculations already used the old value

Solution: Extract the hardcoded 4 into a one-line virtual method _get_mma_inst_tile_k() that subclasses can override.

Use case: Our MoE GEMM kernel (MoeGemmSm100(GemmSm100)) overrides this to return 8 for BF16 (128-element K-tile to amortize per-expert prefix-sum overhead) and 4 for blockscaled MXFP8.

Impact: +8 lines, zero behavioral change, purely additive extensibility point. All existing tests pass unmodified.

### DIFF
--- a/quack/gemm_sm100.py
+++ b/quack/gemm_sm100.py
@@ -302,7 +302,7 @@ class GemmSm100(GemmSm90):
             )
 
         # Compute mma/cluster/tile shapes
-        mma_inst_tile_k = self.mma_tiler[2] // self.mma_inst_shape_mnk[2] if self.mma_tiler[2] > 1 else 4
+        mma_inst_tile_k = cute.size(cute.shape_div(self.mma_tiler[2], self.mma_inst_shape_mnk[2]))
         self.mma_tiler = (
             self.mma_tiler[0],
             self.mma_tiler[1],
--- a/quack/gemm_sm100.py
+++ b/quack/gemm_sm100.py
@@ -240,7 +240,6 @@ class GemmSm100(GemmSm90):
         # Multiple of 4 warps to increase/decrease number of registers
         assert self.threads_per_cta % 128 == 0
 
-
     def _setup_attributes(self, epilogue_args: EpilogueArguments, varlen_args: VarlenArguments):
         """Set up configurations that are dependent on GEMM inputs
 
--- a/quack/gemm_sm100.py
+++ b/quack/gemm_sm100.py
@@ -240,6 +240,7 @@ class GemmSm100(GemmSm90):
         # Multiple of 4 warps to increase/decrease number of registers
         assert self.threads_per_cta % 128 == 0
 
+
     def _setup_attributes(self, epilogue_args: EpilogueArguments, varlen_args: VarlenArguments):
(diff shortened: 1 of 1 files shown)


quack/gemm_sm100.py
```diff
@@ -240,6 +240,14 @@ def __init__(
         # Multiple of 4 warps to increase/decrease number of registers
         assert self.threads_per_cta % 128 == 0
 
+    def _get_mma_inst_tile_k(self):
+        """Number of MMA instruction K-tiles per mainloop iteration.
+
+        Override in subclasses to change the K-tile depth (e.g. MoE uses 8
+        for BF16 to reach a 128-element K-tile).
+        """
+        return 4
```
thakkarV: why is this hard coded to 4? this should be inferred by applying the tiled MMA to the tile shape mnk and then taking the size of the restK instead.

thakkarV:
@tridao have you considered exposing tile shape K as a parameter as well rather than just tile shape mn? that would avoid having to do this.

quack/gemm_sm100.py
santoshmo: changed the hardcoded `4` to `mma_inst_tile_k: Literal[4] | Unknown = self.mma_tiler[2] // self.mma_inst_shape_mnk[2] if self.mma_tiler[2] > 1 else 4`

quack/gemm_sm100.py
```diff
@@ -302,7 +302,7 @@ def _setup_attributes(self, epilogue_args: EpilogueArguments, varlen_args: Varle
             )
 
         # Compute mma/cluster/tile shapes
-        mma_inst_tile_k = 4
+        mma_inst_tile_k = self.mma_tiler[2] // self.mma_inst_shape_mnk[2] if self.mma_tiler[2] > 1 else 4
```
thakkarV: this assumes MMA tiler and instruction shape are flat, which is not always a sound assumption:


```suggestion
        mma_inst_tile_k = cute.size(cute.shape_div(self.mma_tiler[2], self.mma_inst_shape_mnk[2]))
```

also why do you have a conditional path with default of 4 here still?

tridao:
Yeah i think exposing tile_shape_mnk would be the cleanest API. If only 2 values are passed, then we set K = 4 * instruction K shape.

santoshmo:
> Yeah i think exposing tile_shape_mnk would be the cleanest API. If only 2 values are passed, then we set K = 4 * instruction K shape.

@tridao, please take a look

tridao (approved):

PR #103, opened 2026-04-08
Fix gemm_out inplace versus separate out tensor pytorch custom op schema mismatch

thakkarV:
`gemm_add()` had an invalid dispatch path for the out is C case. When callers reused C storage as the output, the wrapper marked the operation as in-place but still called `gemm_add_out()` with `C=None`. That violates the `gemm_add_out` custom-op contract, which requires `C: Tensor`, and can trigger PyTorch schema/assert failures.

This change fixes the wrapper to dispatch that case to `gemm_add_inplace()` instead. The normal out-of-place path still goes through `gemm_add_out()`.

A regression test was added to cover `gemm_add(A, B, C, out=C, alpha=tensor, beta=1.0)` so the wrapper contract stays correct.

### DIFF
--- a/quack/gemm_interface.py
+++ b/quack/gemm_interface.py
@@ -562,23 +562,40 @@ def gemm_add(
     alpha = alpha if isinstance(alpha, float) else 1.0
     beta_tensor = beta if not isinstance(beta, float) else None
     beta = beta if isinstance(beta, float) else 1.0
-    gemm_add_out(
-        A,
-        B,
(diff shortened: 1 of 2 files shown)


thakkarV:
@tridao  is main broken? the failing test is an RMS norm test.

tridao:
hmm let me see, CI was green before, idk why rmsnorm suddenly fails

tridao (approved):

PR #104, opened 2026-04-10
Fix autotuner worker device affinity

NoahAmsel:
I observed the following issue when using quack in combination with fsdp on 8 B200s. All 8 GPUs run quack autotuning simultaneously, in a subprocesses. It seems that all these subprocesses run on cuda:0, the first GPU. They also all take up memory. Unfortunately, this causes an OOM. The other GPUs wait around for their subprocesses to finish, but they never do.

The stack trace showing the OOM error is attached.
[stderr.log.20260410T015022.txt](https://github.com/user-attachments/files/26618493/stderr.log.20260410T015022.txt)

The simplest fix is for each GPU to run its own subprocesses, rather than sending it to cuda:0. This reduces the memory pressure on the first GPU, preventing the OOM. In production, this can make a big difference. I'm using configs that have been designed so that training takes up as much of the GPUs memory as possible without OOMing, so there is not much extra room. The memory overhead of 7 extra processes pushes me over.

This PR simply sets the environment variable CUDA_VISIBLE_DEVICES of the subprocess so that it runs on the current device.

### DIFF
--- a/quack/autotuner.py
+++ b/quack/autotuner.py
@@ -25,6 +25,29 @@ PACKAGE_NAME = "quack"
 VERSION = __version__
 
 
+def _get_current_cuda_device() -> str | None:
+    """Return the physical CUDA device identifier for the current process.
+
+    Maps the logical ``torch.cuda.current_device()`` index through
+    ``CUDA_VISIBLE_DEVICES`` (if set) so the result is valid as a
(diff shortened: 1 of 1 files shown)


tridao:
If CUDA_VISIBLE_DEVICES isn't set, does this assign different GPUs to each subprocess?

tridao (approved):

PR #112, opened 2026-04-19
Fix bias with concat layout

GarlGuo:
We will interleave the bias if the bias has a concat layout. The time regression is minor as bias is small and can fit seamlessly with the current epi tiling strategy.


tridao:
why don't we just change the layout of bias in the __call__ of the kernel?

tridao:
we're using cp.async to load bias, each with 32 bits, so changing the layout should just work

tridao (approved):

PR #116, opened 2026-04-22
[Gemm] Add tile_K parameter and enable 2CTA for tile_m=128

wangyang59:
Add tile_K parameter to gemm() API so callers can set the K tile size (BK) via config instead of relying on the default. When tile_K > 0, it is passed to GemmSm100 to compute mma_inst_tile_k = tile_K / mma_inst_K.

Also enable use_2cta_instrs for tile_m=128. The hardware supports 2CTA for tile_m in {128, 256} but quack only enabled it for tile_m=256. This doubles CTA count for M<=256 shapes, improving SM utilization.

### DIFF
--- a/quack/__init__.py
+++ b/quack/__init__.py
@@ -2,10 +2,17 @@ __version__ = "0.3.11"
 
 import os
 
-from quack.rmsnorm import rmsnorm
-from quack.softmax import softmax
-from quack.cross_entropy import cross_entropy
-from quack.rounding import RoundingMode
+# NVIDIA/cutlass#3161: fix duplicate .text section flags in CuTeDSL-emitted .o
+# files before anything else imports / compiles. Must run before the first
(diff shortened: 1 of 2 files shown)


quack/gemm.py
```diff
@@ -149,6 +149,7 @@ def gemm(
     tile_N: int,
     cluster_M: int,
     cluster_N: int,
+    tile_K: int = 0,
```
thakkarV: make the default None instead of 0?

thakkarV: also, we should add this symmetrically to other GEMM APIs too in case they support tile M and N already

Pranshu-Bahadur:
@wangyang59 Thank you so much! you were right, M=128 2CTA bf16 is legit, I am assuming N is split in multiples of 8 across the 2CTAs which means only 1 CTA needs to have 128xK in tmem right?

PR #117, opened 2026-04-23
[Gemm] Add tile_K parameter and enable 2CTA for tile_m=128

wangyang59:
Add tile_K parameter (default None, SM100/SM110 only) to all GEMM APIs
  so callers can set the K tile size instead of relying on the default.

  Enable use_2cta_instrs for tile_m=128 in GemmSm100. The hardware
  supports 2CTA for tile_m in {128, 256} but only 256 was enabled.


Pranshu-Bahadur:
is this also for bf16 dense mma, not sure is 2CTA bf16 mma is a thing?

wangyang59:
> is this also for bf16 dense mma, not sure is 2CTA bf16 mma is a thing?

yes

quack/gemm_sm100.py
```diff
@@ -188,7 +188,10 @@ def __init__(
         """
 
         self.acc_dtype: Type[cutlass.Numeric] = acc_dtype
-        self.use_2cta_instrs = mma_tiler_mn[0] in (256,)
+        self.use_2cta_instrs = cluster_shape_mnk[0] == 2 and mma_tiler_mn[0] in (
```
tridao: only allow m=128 if not blockscaled

tridao:
Thanks! Whats' the tradeoff between cta=2 vs cta=1 when tile_m=128?

tridao:
the tile_K thing seems fine to me

wangyang59:
> Thanks! Whats' the tradeoff between cta=2 vs cta=1 when tile_m=128?

Here are some shapes we benchmarked. Are there other shapes that you would like to test on? 

Screenshot 2026-04-24 at 12 03 53 PM

quack/gemm_sm100.py
wangyang59: updated

tridao:
Yeah i see, this is the case where there's not enough parallel work so using 2cta would be better

quack/gemm_sm100.py
```diff
@@ -188,16 +188,17 @@ def __init__(
         """
 
         self.acc_dtype: Type[cutlass.Numeric] = acc_dtype
-        self.use_2cta_instrs = mma_tiler_mn[0] in (256,)
+        self.sf_vec_size = sf_vec_size
+        self.blockscaled = sf_vec_size is not None
+        valid_2cta_m = (128, 256) if not self.blockscaled else (256,)
+        self.use_2cta_instrs = cluster_shape_mnk[0] == 2 and mma_tiler_mn[0] in valid_2cta_m
```
tridao: cluster_shape_mnk[0] % 2 == 0

tridao:
One final nit, other than that LGTM. Thank you!

tridao:
Some test failures look real, maybe tile_K wasn't passed properly there?
```
FAILED tests/test_linear.py::test_gemm_norm_act[input_dtype0-4096-2048-False-None-False-True] - ValueError: Expected size in shape to be strictly positive, but got 0
```

Pranshu-Bahadur:
> > Thanks! Whats' the tradeoff between cta=2 vs cta=1 when tile_m=128?
> 
> Here are some shapes we benchmarked. Are there other shapes that you would like to test on?
> 
> Screenshot 2026-04-24 at 12 03 53 PM

@wangyang59 
This is beautiful, thank you so much! 

Is N split across the 2 CTAs in mutiples of 8 (Suppose N=16, Each CTA will have 8 N in their local smem)?
Does this mean only 1 CTA needs to have A matrix in tmem? Or is this split too by K? K/2 in CTA 1 and K/2 in CTA 2?

wangyang59:
> @wangyang59 This is beautiful, thank you so much!
> 
> Is N split across the 2 CTAs in mutiples of 8 (Suppose N=16, Each CTA will have 8 N in their local smem)? Does this mean only 1 CTA needs to have A matrix in tmem? Or is this split too by K? K/2 in CTA 1 and K/2 in CTA 2?

My understanding is that the split is on M-dim for 2-CTA.

wangyang59:
> Some test failures look real, maybe tile_K wasn't passed properly there?
> 
> ```
> FAILED tests/test_linear.py::test_gemm_norm_act[input_dtype0-4096-2048-False-None-False-True] - ValueError: Expected size in shape to be strictly positive, but got 0
> ```

Thanks for raising this. Yeah, it was due to some mismatched argument positions in gemm_inferface.py caused by the added tile_k. Fixed it in the latest version. test_linear.py now passes

tridao (approved):

tridao:
Thank you!

PR #119, opened 2026-04-24
ruff benchmarks/

alecco:


tridao:
can you add it to https://github.com/Dao-AILab/quack/blob/main/.pre-commit-config.yaml

tridao (approved):

PR #118, opened 2026-04-23
Add remaining SM120 / RTX 50 support for GEMM epilogues and RMS paths

alecco:
Author: Alecco (& Codex) for Ologan

## Summary

Add SM120 support for RTX 50 GPUs across missing GEMM epilogue paths and dependent wrappers.

This PR enables and validates:

- gemm_dgated
- non-gated gemm_act
- non-gated gemm_dact
- gemm_sq_reduce
- gemm_rms
- gemm_rms_then_norm_act
- non-gated MLP eager and torch.compile(fullgraph=True)
- varlen non-gated activation and dactivation

It also adds generic GEMM epilogue profiling/benchmark tooling, SM120-specific profiling docs, and an SM120 gemm_rms default selected from repeated local measurements.

NOTE: I took the liberty to list RTX 50 as supported, but I understand this is a repo maintainer’s responsibility and it would entail more than just a drive-by PR. Please
let me know if I can help or if you’d prefer it removed for now.

### Kernel and interface support

- enable SM120 non-gated activation forward in quack/gemm_act.py
- enable SM120 non-gated dactivation backward in quack/gemm_dact.py
- enable SM120 sq-reduce / RMS paths in quack/gemm_sq_reduce.py
- add an SM120-specific gemm_rms default in quack/gemm_interface.py
- keep autotune pruning correctness-only rather than pruning valid configs for local RTX 5060 performance

### ColVecReduce fix

- generalize ColVecReduce in quack/epi_ops.py to support inter-warp reduction when warp_N > 1
- preserve the old fast path for warp_N == 1
- this fixes the structural blocker behind SM120 sq-reduce / RMS

### Wrapper-level compile fix

- fix torch.compile(fullgraph=True) wrapper behavior in quack/linear.py and quack/mlp.py
- avoid tracing method dispatch through ops classes inside autograd wrapper code
- store concrete callables on the autograd context instead
- this is needed for non-gated MLP compile coverage on the local PyTorch 2.11.0+cu130 stack

### GEMM epilogue profiling and benchmark tooling

- add benchmarks/benchmark_gemm_epilogues.py as a generic GEMM epilogue benchmark harness for:
    - gemm_rms
    - gemm_norm_act
    - gemm_act
- support explicit config overrides from the CLI:
    - tile_m
    - tile_n
    - pingpong
    - swap_ab
    - dynamic persistence mode
- explicit configs use the active GPU capability instead of hardcoding SM120
- support benchmarking selected/default paths with:
    - --use-selected-config
[...]

PR #120, opened 2026-04-25
Fix ColVecReduce shared-memory staging

alecco:
My https://github.com/Dao-AILab/quack/pull/118 added shared-memory staging to ColVecReduce so SM120 epilogues with warps_in_N > 1 can reduce partial row sums across N-warps.

This PR fixes two issues:

1. The staging buffer was allocated for all architectures, even though SM90/SM100/SM110 use the original warps_in_N == 1 direct-write path.
2. SM120 reserved space for 8 N-warps, which was overly conservative.

New allocation policy:

SM90/SM100/SM110: 0 bytes
SM120:            tile_M * 2 * sizeof(float32)

For the default SM120 RMS config with tile_M=64, this reduces extra staging from:

64 * 8 * 4 = 2048B

to:

64 * 2 * 4 = 512B

## Why 2 N-warps on SM120

I temporarily instrumented ColVecReduce.end() after layout analysis:

lane_layout_MN, warp_layout_MN = _get_lane_warp_layouts(tiled_copy, reference_src)
lanes_in_N = cute.size(lane_layout_MN, mode=[1])
warps_in_N = cute.size(warp_layout_MN[1])

I checked representative SM120 RMS configs:

tile_m=64  tile_n=128 pingpong=True
tile_m=128 tile_n=64  pingpong=False
tile_m=128 tile_n=128 pingpong=False
tile_m=128 tile_n=160 pingpong=False
tile_m=128 tile_n=192 pingpong=False
tile_m=128 tile_n=64  pingpong=True
tile_m=128 tile_n=128 pingpong=True
tile_m=128 tile_n=160 pingpong=True

All reported:

lanes_in_N=4
warps_in_N=2

So SM120 only needs one FP32 partial per M row per two N-warps.

For non-SM120, the PR keeps zero extra shared memory and asserts if a future layout unexpectedly reports warps_in_N > 1.

## Validation

Added tests/test_epi_ops.py to verify:

SM90/SM100/SM110 -> 0 bytes
SM120            -> tile_M * 2 * sizeof(float32)
missing tensor   -> 0 bytes

Also ran an SM120 gemm_rms correctness check with the reduced staging.

### DIFF
--- a/quack/epi_ops.py
+++ b/quack/epi_ops.py
@@ -545,7 +545,9 @@ class ColVecReduce(EpiOp):
     and final reduction + gmem write (end).
     """
 
-    max_warps_in_n = 8
+    def __init__(self, name, max_warps_in_n=1):
+        super().__init__(name)
+        self.max_warps_in_n = max_warps_in_n
 
     def param_fields(self):
         return [(self.name, object, None)]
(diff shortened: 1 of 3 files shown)


quack/epi_ops.py
```diff
@@ -545,29 +545,40 @@ class ColVecReduce(EpiOp):
     and final reduction + gmem write (end).
     """
 
-    max_warps_in_n = 8
+    # SM120 warp-MMA epilogues observed in RMS configs split each row's N
```
tridao: we shouldn't pass `arch` around. We should just pass warps_in_n in the constructor of this EpiOps

tridao: conceptually, nothing prevents hopper gemm to use 2 warps_in_n either (though in practice we don't do that).

quack/epi_ops.py
```diff
@@ -552,7 +545,9 @@ class ColVecReduce(EpiOp):
     and final reduction + gmem write (end).
     """
 
-    max_warps_in_n = 8
+    def __init__(self, name, max_warps_in_n=0):
```
tridao: default shoould be 1

quack/epi_ops.py
```diff
@@ -567,11 +562,13 @@ def smem_bytes(self, arg_tensor, cta_tile_shape_mnk, epi_tile):
 
     def smem_struct_field(self, gemm, params):
         tensor = getattr(params, self.name, None)
-        size = gemm.cta_tile_shape_mnk[0] * self.max_warps_in_n if tensor is not None else 0
+        if tensor is None or self.max_warps_in_n == 0:
```
tridao: == 1

tridao (approved):

tridao:
Thank you !

PR #127, opened 2026-04-30
Add Sm120 blockscaled FP4 GEMM path

alecco:
### Requires https://github.com/NVIDIA/cutlass/pull/3185

Author: Alecco (& Codex) for Ologan

## Summary

This PR adds the first guarded SM120 blockscaled GEMM path using the existing blockscaled GEMM interface, rather than adding a separate SM120-specific frontend.

**This is a first implementation _not optimized_**. Optimization will be done in a subsequent PR.

Supported SM120 scope:

- A/B: Float4E2M1FN FP4 operands, packed as torch.float4_e2m1fn_x2
- NVFP4 scales: Float8E4M3FN, sf_vec_size=16
- MXFP4 scales: Float8E8M0FNU, sf_vec_size=32
- Accumulator: Float32
- Output: BFloat16
- Shapes: M % 128 == 0, N % 128 == 0, K % 64 == 0, L == 1
- Tile/cluster: tile_shape_mnk=(128,128,64), cluster_shape_mnk=(1,1,1)

Unsupported for this first path:

- C / beta
- varlen
- gather_A
- grouped/sparse kernels
- multicast clusters
- output FP4 quantization
- generic autodispatch
- SM120 FP8/INT8 blockscaled paths

## Implementation Notes

SM120 uses a padded row-major physical scale layout:

logical_scale_cols = ceil(K / sf_vec_size)
physical_scale_cols = round_up(logical_scale_cols, 16)

SFA shape = (M, physical_scale_cols, 1)
SFB shape = (N, physical_scale_cols, 1)

Compact scale rows such as (M, 4) are rejected before launch because SM120 TMA on compact row-major scale pages with row stride below 16 bytes traps. Padding columns are
ignored by the kernel.

For K > 64, K64 partials are accumulated through an FP32 shared-memory scratch tile and D is converted to BF16 only once on the final K tile. This preserves FP32
accumulation semantics instead of rounding through BF16 between K tiles.

The SM120 FFI compile path still calls the JIT-only blockscaled entry directly, after host-side validation, because it compiles logical fake CuTe tensor views over packed
torch storage. The class-level blockscaled_call keeps host validation for logical class-call tensors.

The benchmark path can launch the SM120 blockscaled kernel and build padded scale tensors, but cuBLAS comparison is skipped because the benchmark currently builds QuACK’s
padded row-major scale layout, not the cuBLAS/PyTorch scaled_mm scale layout.

## Tests

Validated with:

python -m compileall quack tests benchmarks
ruff check quack tests benchmarks
CUTE_DSL_ARCH=sm_120a pytest -q tests/test_gemm_blockscaled.py -s -rs
pytest -q 'tests/test_linear.py::test_gemm[960-736-1504-input_dtype0-k-k-m-False]' -s -rs

Current results on SM120:

- tests/test_gemm_blockscaled.py: 22 passed, 52 skipped
- Dense GEMM smoke: 1 passed

The SM120 tests cover:

- supported/rejected can_implement_blockscaled cases
- padded scale layout validation
- compact scale rejection
- direct class-call validation
- scale-sensitive runtime correctness
- K=320 scale-page crossing
- K=384 regression for FP32 accumulation across K64 tiles
- non-constant FP4 K-lane patterns for A/B with scale padding poison present

## Notes

This PR does not claim full arbitrary FP4 operand-pattern coverage yet. Broader row-and-K-varying FP4 correctness, cuBLAS-compatible SM120 scale conversion, and additional
blockscaled features should be follow-up work.

## Agent disclosure

This work was created with OpenAI Codex CLI agent.

### DIFF
--- a/AI/varlen_blockscaled_sf_layout.md
+++ b/AI/varlen_blockscaled_sf_layout.md
@@ -152,7 +152,7 @@ a tile-unit offset integer.
 
 ## Tests
 
-`tests/test_gemm_sm100_blockscaled.py`:
+`tests/test_gemm_blockscaled.py`:
 - `test_blockscaled_mxfp8_varlen_m_nonaligned` — 4 seqlen patterns × 2 B-majors = 8 cases.
   Patterns include `[128, 128, 128]`, `[100, 200, 150]`, `[30, 300, 64, 200]`,
   `[1, 128, 127, 129]`.
--- a/benchmarks/benchmark_gemm.py
+++ b/benchmarks/benchmark_gemm.py
@@ -194,19 +194,21 @@ def _run_blockscaled(args):
         compile_blockscaled_gemm_tvm_ffi,
         create_blockscaled_operand_quantized,
         create_blockscaled_operand_tensor,
+        create_sm120_blockscaled_scale_tensor,
         create_blockscaled_varlen_m_operands,
         scale_blocked_for_cublas,
         torch_dtype_for_cutlass,
     )
     from quack.cute_dsl_utils import get_device_capacity
-    from quack.gemm_default_epi import GemmDefaultSm100
+    from quack.gemm_default_epi import GemmDefaultSm100, GemmDefaultSm120
 
     sm_major = get_device_capacity(torch.device("cuda"))[0]
-    assert sm_major in (10, 11), (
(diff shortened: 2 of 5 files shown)


alecco:
Fixed benchmark BF16 checks and docstring.

PR #129, opened 2026-05-01
[minor] pytest asyncio default

alecco:
To avoid a lot of warnings, set asyncio_mode aut (as recommended by the warning).


tridao:
Quack doesn't depend on pytest-asyncio and has no async tests, so this config doesn't apply.

PR #131, opened 2026-05-06
fix the SM100 2CTA issue for SwiGLU/dSwiGLU

GarlGuo:
Claude gave a minimum reproduction script. After this fix, the original discrepancy on swiglu results appears to be fixed. 

And the tests are all passed. 
```python
"""Minimal A/B test for the gated SM100 fix.

Phase A — _valid_2cta_m returns (128, 256): the original buggy default.
Phase B — _valid_2cta_m returns (256,):       the patched default.

Each phase runs in its own subprocess with its own QUACK_CACHE_DIR so the
disk-backed jit_cache (whose key doesn't include use_2cta_instrs) can't serve
a stale cross-phase kernel.

Usage: CUDA_VISIBLE_DEVICES=<gpu> python solo_ab_min.py
"""
import json, os, shutil, subprocess, sys, tempfile, torch


def child(phase):
    valid = (128, 256) if phase == "A" else (256,)
    from quack.gemm_act import GemmGatedMixin
    GemmGatedMixin._valid_2cta_m = lambda self, _v=valid: _v

    from quack.gemm_config import GemmConfig
    from quack.gemm_interface import gemm_gated_tuned, gemm_gated_ref

    M, H, I, E = 32768, 1024, 512, 8
    dtype, dev = torch.float16, torch.device("cuda:0")
    g = torch.Generator(device=dev).manual_seed(0)
    counts = torch.full((E,), M // E, dtype=torch.int32, device=dev)
    cu = torch.zeros(E + 1, dtype=torch.int32, device=dev)
    cu[1:] = torch.cumsum(counts, dim=0).to(torch.int32)
    T = M // 4
    x = (0.02 * torch.randn(T, H, generator=g, device=dev, dtype=torch.float32)).to(dtype)
    A_idx = torch.randint(0, T, (M,), dtype=torch.int32, device=dev, generator=g)
    w = torch.empty(E, 2 * I, H, dtype=torch.float32, device=dev)
    torch.nn.init.normal_(w, mean=0.0, std=0.02, generator=g)
    w1 = w.to(dtype).permute(1, 2, 0).permute(2, 1, 0)

    cfg = GemmConfig(tile_m=128, tile_n=256, cluster_m=2, cluster_n=1,
                     swap_ab=False, max_swizzle_size=8,
                     is_dynamic_persistent=True, use_tma_gather=True,
                     pingpong=False, device_capacity=10)
    pre, post = torch.empty(M, 2 * I, dtype=dtype, device=dev), torch.empty(M, I, dtype=dtype, device=dev)
    gemm_gated_tuned.fn(x, w1, pre, post, None, None, "swiglu", cu, A_idx, False, config=cfg)
    torch.cuda.synchronize()
    pre_ref, post_ref = gemm_gated_ref(x, w1, bias=None, activation="swiglu",
                                       cu_seqlens_m=cu, A_idx=A_idx,
                                       store_preact=True, concat_layout=None)

    print("PHASE_RESULT " + json.dumps({
        "phase": phase, "valid_2cta_m": list(valid),
        "preact_max_abs": (pre.float() - pre_ref.float()).abs().max().item(),
        "preact_max_ref": pre_ref.float().abs().max().item(),
        "postact_max_abs": (post.float() - post_ref.float()).abs().max().item(),
        "postact_max_ref": post_ref.float().abs().max().item(),
    }), flush=True)


def run_phase(phase):
    cache = tempfile.mkdtemp(prefix=f"quack_cache_{phase}_")
    env = {**os.environ, "QUACK_CACHE_DIR": cache}
    try:
        out = subprocess.run([sys.executable, "-u", __file__, phase],
                             capture_output=True, text=True, env=env, timeout=600)
    finally:
        shutil.rmtree(cache, ignore_errors=True)
    for line in out.stdout.splitlines():
        if line.startswith("PHASE_RESULT "):
            return json.loads(line[len("PHASE_RESULT "):])
    sys.exit(f"phase {phase} produced no result\n{out.stdout}\n{out.stderr}")


def main():
    if len(sys.argv) > 1 and sys.argv[1] in ("A", "B"):
        child(sys.argv[1]); return

    a, b = run_phase("A"), run_phase("B")
    for d in (a, b):
        rel = d["preact_max_abs"] / max(d["preact_max_ref"], 1e-12)
        print(f"phase {d['phase']}  _valid_2cta_m={str(tuple(d['valid_2cta_m'])):<10}  "
              f"preact rel={rel:.4e}  ({'FAIL' if rel > 0.05 else 'PASS'})")


if __name__ == "__main__":
    main()
```

### DIFF
--- a/AI/complex64_design_notes.md
+++ b/AI/complex64_design_notes.md
@@ -1,0 +1,255 @@
+# `quack.complex.Complex64` design notes
+
+A native-feeling `complex64` element type for CuTe-DSL kernels, without
+patching the C++ MLIR bindings. Lives in `quack/complex.py`. Validated by
+`tests/test_complex.py` (44 tests covering constructors, arithmetic matrix,
+methods, class invariants, helpers, the c_pointers boundary, special / edge
+values, auto-promotion, and a smem round-trip). FFT-correctness tests that
+exercise Complex64 through real kernels live in `tests/test_fft.py`.
+
+This file records *why* the design looks the way it does and what was tried
+and rejected. The production native C2C FFT path in `quack/fft.py` already
+uses Complex64 for register values and IO -- see
+`AI/complex64_fft_migration_plan.md` for the cutover history and remaining
+work.
+
+---
+
+## The fundamental constraint
+
+`cute.MemRefType.get(value_type, layout_type)` -- the C++ MLIR-binding call
+that backs every `make_tensor`, `make_rmem_tensor`, `recast_tensor`, and smem
+`allocate_tensor` -- enforces an element-type **allowlist**: the value_type
+must be one of `int`, `float`, `ptr`, or `sparse_elem`. The natural choice
+for "complex single-precision" -- `complex<f32>` -- gets:
+
+```
+TypeError: expects value type to be int, float, ptr or sparse_elem,
+           but got 'complex<f32>'
+```
+
+This is enforced in `_cute_ir` (the compiled binding), not in Python. We
+have no Python-level escape hatch.
+
+The other plausible "unique" MLIR types we tried have the same fate:
+
+| MLIR element type     | allowlisted? | unique to Complex64?              |
+| --------------------- | :----------: | --------------------------------- |
+| `complex<f32>`        | ❌            | unique but rejected               |
+| `tuple<f32, f32>`     | ❌            | unique but rejected               |
+| `f64`                 | ✅            | shared with `Float64`             |
+| `i64`                 | ✅            | shared with `Int64` / `Uint64`    |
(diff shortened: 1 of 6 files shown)


tridao:
I see. The fix would work but i'll try to see if we can generalize the epilogue code to deal with (2,2) warp anyway. That would be cleaner, instead of papering over it.
And we should add a test for this (2,2) case!

GarlGuo:
@tridao I can check to see if this is implementable.

GarlGuo:
#133 will propose another fix

GarlGuo:
not needed as #133 has been merged

PR #132, opened 2026-05-06
RMS Norm Bwd: Optimize for GB200 w/ TMA loads, reload_x, separate output dtype, new launch configs

thakkarV:
Various optimizations and enhancements to RMSNormBwd kernel for GB200

## Kernel additions

- TMA 1D or 2D loads for X / dY load path depending on the CTA tilers. Helps reduce register pressure for larger reduction dims without having to bump up the cluster shapes.
- Refactored persistent-loop loads: refactors the load loop to allow `consumer_state.index` (TMA) and `stage` (cp.async), share the same compute body.
- TMA uses the same smem stage count as LDGSTS.
- Runtime TMA-eligibility guard: `USE_TMA = config.use_tma and not per_head and row_bytes_x % 16 == 0 and row_bytes_do % 16 == 0` - silently falls back to cp.async when row bytes aren't 16-B aligned.
- `reload_x="smem"` knob mirroring the existing `reload_wdy`: after the row reduction barrier, X is reloaded from sX and `x_hat` recomputed, freeing the X registers that would otherwise sit live across the barrier.
- Configurable smem stage count: `n_smem_stages = self._tma_stages_val if USE_TMA else 2`, lifting the previously hardcoded 2-stage layout, and bumping smem alignment to 128 B when TMA is enabled.
- Prevent W fragment rematerialization: a no-op fp32 round-trip (`tXrW.store((tXrW.load().to(Float32) + Float32(0.0)).to(tXrW.element_type))`) after the gmem load - value-preserving, but the SSA boundary forces the compiler to keep tXrW register-resident across the persistent loop instead of re-emitting the gmem load on each iteration.

## New launch config class and GB200 specific heuristics

- `T_hint`: row count drives the heuristic but we keep the kernel body's `M` as a runtime symbol so the kernel itself isn't templated on every T value. `T_hint` enters only the `_compile_rmsnorm_bwd` cache key ; adjacent T values that hash to the same config bucket share a compiled binary; T values that pick different config buckets get separate compiles.
- `get_sm_count` lifted out of `rmsnorm.py`: arch-aware (Hopper vs Blackwell) wave-count formulas now live alongside the launch config in the new module.
- Adds separate output dtype to the class ctor for better heuristics selection.
- Config module - `quack/rmsnorm_config.py` mirrors `quack.gemm_config`. A frozen `RmsNormConfig` dataclass holds every launch knob; `RmsNormConfig.select(N, dtype_width, dout_width, arch_major, T_hint)` dispatches to `_for_blackwell` / `_for_hopper`. The Blackwell rule is a single byte-driven ladder over `row_bytes = N * max(x_bytes, dout_bytes)`:

```
row_bytes <= 16K       -> cluster_n=1, layout 128/128
16K < row_bytes < 48K  -> cluster_n=1, layout 256/256
row_bytes => 48K       -> cluster across CTAs:
                          cluster_n=4 if 0 < T <= 1024 AND row_bytes <= 64K
                          cluster_n=8 otherwise

use_tma    = cn=>4  OR  (max_bytes => 4 AND row_bytes => 16K)
reload_x   = "smem" if (cn=>8 AND N => 32K)
                        OR (cn=1 AND max_bytes => 4 AND bytes_per_thread_frag => 64)
reload_wdy = "smem" if cn=>4 OR bytes_per_thread_frag => 64
```

## Benchmarks on GB200

Measured on a single GB200 (NVIDIA HBM3e) across 256 shapes - `T x D in {1K, 2K, 4K, 8K, 12K, 16K, 24K, 32K}` x 4 dtype combos `{bf16/bf16, fp32/fp32, bf16/fp32, fp32/bf16}`.

Each shape captures 10 CUDA graphs over rotating input sets (graph count auto-trimmed to fit GPU memory at the largest shapes - 4 graphs for fp32 32Kx32K) and replays them in round-robin for 500 iters per shape, back to-back. The following heatmap shows the speedup distribution across the four input/output dtype combinations:

image

- Geomean: **+19.7%** speedup vs `quack/master`
- Median: **+12.7%**
- Max: **+90.2%** (T=32K, D=12K, bf16/bf16)
- Min: **-17.1%** (T=24K, D=2K, fp32/fp32)
- Overall: 193 / 256 shapes faster, 13 shapes regressing > 10%, 14 shapes regressing 5–10%

Per-dtype geomean: bf16/bf16 **+28.5%**, fp32/bf16 **+21.9%** (0 regressions), bf16/fp32 **+17.0%**, fp32/fp32 **+12.0%**. The 13 regressions > 10% concentrate in three clear corners of the grid:

- **fp32/fp32 at D=2048**: 5 cells (T ∈ {8K..32K}) regress 15–17%. Same small-N pure-fp32 corner flagged before — the rewritten kernel body has measurable per-call overhead that no heuristic config can recover; master's simpler bwd kernel runs ~15% faster at this T-N corner regardless.
- **bf16/bf16 at D=4096**: 6 cells (T ∈ {4K..32K}) regress 10–13%. New regression band — master's bf16/bf16 kernel for this medium-N row was tuned more tightly than the new heuristic ladder lands on. This is the most likely first target if we extend the analytical heuristic.
- **bf16/fp32 at D=16384, T=2048**: 1 isolated cell at −15%.

fp32/bf16 has zero regressions across all 64 of its shapes; bf16/bf16 still posts the biggest single win at **+90.2%** (T=32K, D=12K).

### %SOL utilization before and after

The same 256-shape grid expressed as **% of GB200 HBM3e peak (8 TB/s per GPU)**, on a fixed 0–100 RdYlGn scale so the two plots are directly comparable cell-by-cell.

quack master:

image

| dtype combo | geomean %SOL | peak %SOL |
|---|---|---|
| bf16/bf16 | 33 | 62 |
| fp32/fp32 | 52 | 81 |
| bf16/fp32 | 41 | 74 |
| fp32/bf16 | 45 | 74 |

This MR:

image

| dtype combo | geomean %SOL | peak %SOL |
|---|---|---|
| bf16/bf16 | 42 | 82 |
| fp32/fp32 | 58 | 88 |
| bf16/fp32 | 49 | 83 |
| fp32/bf16 | 55 | 85 |

On master only fp32/fp32 ever crosses 80% SOL, and only in its top-right corner; the other three dtype combos cap at ~74% even at the largest shapes. This MR pushes every dtype combo above 80% SOL at its peak (82–88%) and widens the saturated band to most of D≥8K, T≥8K. The bottom rows (D≤2K) stay red on both plots — they remain latency-bound.

## TODOs for future MRs:

- Port TMA usage to the fwd kernel and see if it helps there?
- Port the `RMSNormConfig` to the fwd kernel too and use it consistently
- Implement `gemm_tuned()` like autotuning for the RMS norm kernels (and all other reduction like kernels too)
- See if the regressions at FP32→FP32 D=2048 and at the new BF16→BF16 D=4096 band are worth extending the default analytical launch configs for.

### DIFF
--- a/quack/rmsnorm.py
+++ b/quack/rmsnorm.py
@@ -8,7 +8,9 @@ import cuda.bindings.driver as cuda
 
 import cutlass
 import cutlass.cute as cute
-from cutlass import Float32, Int32, const_expr
+import cutlass.pipeline as pipeline
+from cutlass import Float32, Int32, Int64, const_expr
+from cutlass.cute.nvgpu import cpasync
 
 import torch
 from torch import Tensor
@@ -22,6 +24,7 @@ from quack.reduce import row_reduce
 from quack.reduction_base import ReductionBase
 from quack.cache_utils import jit_cache
 from quack.cute_dsl_utils import torch2cute_dtype_map
+from quack.rmsnorm_config import RmsNormConfig, get_sm_count
 from cutlass.base_dsl.arch import Arch
 
 
@@ -502,23 +505,47 @@ def rmsnorm_bwd_ref(x, w, dout, rstd, eps=1e-6):
 
 
 class RMSNormBackward(ReductionBase):
-    def __init__(self, dtype: cutlass.Numeric, N: int):
+    def __init__(
+        self,
+        dtype: cutlass.Numeric,
+        N: int,
+        dout_dtype: Optional[Type[cutlass.Numeric]] = None,
+        T_hint: int = 0,
+        per_head: bool = False,
+    ):
         # 2 stages for double buffering when computing mean of x_hat * wdy
         super().__init__(dtype, N, stage=2, reduction_dtype=Float32)
-        self.reload_wdy = None if N <= 16 * 1024 else "smem"
+        dout_width = dout_dtype.width if dout_dtype is not None else dtype.width
+        arch_major = (
+            torch.cuda.get_device_capability(torch.cuda.current_device())[0]
+            if torch.cuda.is_available()
+            else 0
+        )
+        config = RmsNormConfig.select(N, dtype.width, dout_width, arch_major, T_hint)
+        self.config = config
+        self.reload_wdy = config.reload_wdy
+        self.reload_x = config.reload_x
+        self.per_head = per_head
+        self._num_threads_val = config.num_threads
+        self._threads_per_row_val = config.threads_per_row
+        self._cluster_n_val = config.cluster_n
+        self._tma_stages_val = config.tma_stages
+        tile_n = N // max(1, config.cluster_n)
+        row_bytes_x = tile_n * dtype.width // 8
+        row_bytes_do = tile_n * dout_width // 8
+        self.USE_TMA = (
+            config.use_tma and not per_head and row_bytes_x % 16 == 0 and row_bytes_do % 16 == 0
+        )
+        self._can_use_tma = self.USE_TMA
         if self.N > 128 * 1024 and self.dtype.width >= 32:
             # Not enough smem
             raise ValueError("RMSNormBackward does not support N > 128k with dtype >= 32 bits")
 
     def _num_threads(self):
-        return 128 if self.N <= 4096 else 256
+        return self._num_threads_val
 
     def _threads_per_row(self):
-        N = self.N
-        for limit, threads in [(64, 8), (128, 16), (256, 32), (512, 64), (4096, 128)]:
-            if N <= limit:
-                return threads
-        return 256
+        return self._threads_per_row_val
 
     def _set_cluster_n(self):
         arch = cutlass.base_dsl.BaseDSL._get_dsl().get_arch_enum()
@@ -526,19 +553,8 @@ class RMSNormBackward(ReductionBase):
         if arch < Arch.sm_90:
             self.cluster_n = 1
             return
-        # SM12x supports cluster up to 8
         max_cluster = 8 if arch.major == 12 else 16
-        N = self.N
-        if arch.major == 12 and const_expr(self.dtype.width >= 32):
-            # SM12x 99 KB SMEM: fp32 bwd double-buffers 2 tensors, needs much tighter clustering
-            thresholds = [(1024, 1), (8 * 1024, 2), (16 * 1024, 4), (32 * 1024, 8)]
-        else:
-            thresholds = [(8 * 1024, 1), (16 * 1024, 2), (32 * 1024, 4), (64 * 1024, 8)]
-        for limit, cluster in thresholds:
-            if N <= limit:
-                self.cluster_n = cluster
-                return
-        self.cluster_n = max_cluster
+        self.cluster_n = min(self._cluster_n_val, max_cluster)
 
     @cute.jit
     def __call__(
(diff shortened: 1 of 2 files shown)


GarlGuo:
@thakkarV @tridao do you think autotuning will help here? I am thinking of wiring autotuner with RMSNorm/Softmax/Cross entropy.

thakkarV:
> @thakkarV @tridao do you think autotuning will help here? I am thinking of wiring autotuner with RMSNorm/Softmax/Cross entropy. 

Yes it is already listed as one of the TODOs for the future. This is exactly why I added the config class support too - despite my best attempts it is nearly impossible to come up with a good analytical launch config function for these kernels and the regret against best autotuned config can be massive in some cases.

Pranshu-Bahadur:
> Various optimizations and enhancements to RMSNormBwd kernel for GB200
> ...
> - See if the somewhat large regressions for FP32->FP32 @ D=2k are worth extending the default analytical launch configs for.

Hey @thakkarV, 

thank you I'm learning a lot from this. I agree with your insight about TMA, was relieved to see a reasoned out interpretation. 

I have a naive question, when you say cp.async for <16B for gmem to smem loads, I'm assuming you're referring to the cp.async (sm80 era) variant and unless your insight is at <16B both cp.async and cp.async.bulk (normal) loads behave similarly, would make sense from what I understand about gmem loads needing to be multiple of 32-B across a warp to be coalesced (CUDA Programming Guide 13.2 pg44).

Thank you, sorry if the query is off track, I have a lot to learn, any pointers would be amazing 🙏. I am used to CUDA C++, and am still learning the python dsl syntax.

thakkarV:
@Pranshu-Bahadur yes I am referring to LDGSTS, not bulk copy. the vectorization width has to be reduced if the strides are not aligned. LGSTS supports 32b loads so we drop to that in case strides are not 128b aligned. This is orthogonal to coalescing

Pranshu-Bahadur:
> @Pranshu-Bahadur yes I am referring to LDGSTS, not bulk copy. the vectorization width has to be reduced if the strides are not aligned. LGSTS supports 32b loads so we drop to that in case strides are not 128b aligned. This is orthogonal to coalescing 

@thakkarV Oh I see, makes sense thank you so much for the clarification! 

This is very interesting to me, so essentially you maximize the L1 cache use this way leveraging ACCESS mode, acc to CUDA Programming Guide (CPG), what's interesting the PTX ISA 9.2's cp.async variant allows .ca qualifier even with 16B (128b), however the CPG says LDGSTS skips L1 on 16B loads as it triggers BYPASS mode (afaik), this is equivalent to .cg in ptx.

Thank you so much, for your response, this has been very insightful.

quack/rmsnorm.py
```diff
@@ -692,52 +751,165 @@ def kernel(
...
-            copy(tXgX[None, None, None, bidx_start], tXsX[None, None, None, 0], is_async=True)
-            copy(tXgdO[None, None, None, bidx_start], tXsdO[None, None, None, 0], is_async=True)
-        else:
-            if const_expr(tiler_mn[0] > 1):
-                # Fill with zero, otherwise smem will be uninitialized, and we could read this back
-                # later into registers, causing wrong dW.
-                utils.fill_oob(tXsX[None, None, None, 0], None, fill_value=mX.element_type.zero)
-                utils.fill_oob(tXsdO[None, None, None, 0], None, fill_value=mdO.element_type.zero)
-        cute.arch.cp_async_commit_group()
+            # No-op fp32 round-trip; pins tXrW into stable registers across the loop.
```
tridao, line 769: can you add the context in the PR here. This comment is currently pretty cryptic.

tridao (approved):

tridao:
Just checking if T_hint is binned into buckets so that we don't have to recompile for every single value of M?

tridao:
@thakkarV Let's merge when you think it's ready (and T_hint is binned into buckets)

thakkarV:
> @thakkarV Let's merge when you think it's ready (and T_hint is binned into buckets)

Sounds good. Will do this change soon (before EoW)

PR #137, opened 2026-05-14
Fix exception chaining in CuTeDSL ELF patch

jaydenC88:
## Summary
- Chain the caught exception when wrapping object-file read failures in DSLRuntimeError
- Keeps the original OS error available to debuggers and satisfies Ruff B904

## Test
- ruff check --select B904 quack/cute_dsl_elf_fix.py

### DIFF
--- a/quack/cute_dsl_elf_fix.py
+++ b/quack/cute_dsl_elf_fix.py
@@ -74,7 +74,7 @@ def patch() -> None:
                 with open(file_path, "rb") as f:
                     object_file_content = f.read()
             except Exception as e:
-                raise DSLRuntimeError(f"Failed to read object file {file_path}: {e}")
+                raise DSLRuntimeError(f"Failed to read object file {file_path}: {e}") from e
 
         useJitLink = not enable_tvm_ffi
         if not useJitLink and object_file_content:


jaydenC88:
Some context for why this one-liner matters beyond the local B904 hit:

PyTorch vendors this file at `torch/_vendor/quack/cute_dsl_elf_fix.py` and runs Ruff `B904` over the whole tree, so the unchained `raise DSLRuntimeError(...)` surfaces as a lint failure downstream (pytorch/pytorch#183394). PyTorch's policy is *"we don't edit vendored code, please fix upstream"*, so the only correct place to land the fix is here — once merged, the next revendor pulls it into PyTorch automatically and there's no in-tree patch to maintain.

Adding `from e` is also strictly better on its own merits: it preserves the underlying `OSError` (file-not-found, permission, EIO, …) in `__cause__` instead of shadowing it behind the wrapper, which makes the read-failure path much easier to diagnose.

jaydenC88:
@tridao Could you review this PR when you have a chance?

tridao (approved):

tridao:
Thanks! Should we align the ruff rules (e.g. adding B904) to avoid similar issue in the future?

jaydenC88:
> Thanks! Should we align the ruff rules (e.g. adding B904) to avoid similar issue in the future?

I agree. PR is created here https://github.com/Dao-AILab/quack/pull/140

PR #138, opened 2026-05-14
[RMSNorm] Support autotuning of RMSNorm{Fwd|Bwd} via config classes

thakkarV:
This MR adds quack's `gemm_tuned`-style `@autotune`-ing support for both RMSNorm fwd and bwd. The existing `rmsnorm_{fwd,bwd}` analytical-heuristic paths are unchanged; the new `rmsnorm_{fwd,bwd}_tuned` paths bench an exhaustive launch-config search space at first call for each (shape, dtype, key) and cache the winner to disk for subsequent calls.

## Config class plumbing

- New `RmsNormFwdConfig` frozen dataclass mirroring `RmsNormBwdConfig`. Captures every fwd launch knob.
- The fwd heuristic that previously lived inline in RMSNorm is now factored into `RmsNormFwdConfig.from_analytical_heuristic()` + `_for_hopper_fwd()` (Hopper/Blackwell/SM12x share the same ladder for now; a `_for_blackwell_fwd` is easy to add later).
- `RMSNorm.__init__` and `RMSNormBackward.__init__` accept an optional `config` parameter. When `None` they fall back to the analytical heuristic so existing call sites are unaffected.
- `_compile_rmsnorm_{fwd,bwd}` accept an optional config parameter and plumb it through to the kernel class with config captured in the JIT cache key so each chosen config compiles once.

## Search space and pruning

- `get_all_fwd_configs()` and `get_all_bwd_configs()` enumerate the launch-knob product, bounded by the current device's max cluster size and TMA availability:
  - **Fwd**: `num_threads in {128, 256}` x `threads_per_row in {64, 128, 256}` x `cluster_n in {1..max}` x `reload_from in {None, "smem", "gmem"}` ~ **75 configs**
  - **Bwd**: same `num_threads`/`threads_per_row`/`cluster_n` plus `reload_wdy, reload_x in {None, "smem"}` x `use_tma in {False, True}` (SM90+ only) x `smem_stages in {2, 3}` ~ **400 configs**
- Tiny `threads_per_row in {8, 16, 32}` and `smem_stages = 4` are kept in the analytical heuristic (for the narrow-row safety floor `N <= 512`) but trimmed from the autotune search space - stages=4 doubles the data-buffer footprint vs stages=2 and crashes for fp32 N>=64K on Blackwell (227 KB opt-in smem); the tiny widths are only optimal for the narrow-row corner that's not interesting to autotune.
- Both spaces deduplicate at construction (drop `threads_per_row > num_threads` and non-divisible CTA layouts).
- `prune_invalid_rmsnorm_{fwd,bwd}_configs` run as the `@autotune` early-prune step per call:
  - `threads_per_row * cluster_n > N` (over-clusters the row)
  - `use_tma=True` when `per_head=True` (the runtime `USE_TMA` guard would silently fall back, producing identical-timing duplicates in the bench).

## Interface dispatch

New top-level `rmsnorm_fwd_tuned` and `rmsnorm_bwd_tuned` entry points, each decorated with `@autotune`:

- Cache `key=` is `["is_layernorm", "per_head"]` for fwd and `["per_head", "has_dw_partial", "has_db_partial"]` for bwd. The autotuner auto-appends each tensor's shape, dtype, and abstracted stride to the key - adjacent N values produce distinct cache entries; same-shape repeat calls hit the cache.
- `cache_results=True` (default) persists the chosen config to `$QUACK_CACHE_DIR`.
- Calling `.fn(config=...)` bypasses the autotune dispatch entirely.

## Autotuner bench fix (`quack/autotuner.py`)

The default `triton.testing.do_bench` measures kernels with a single input tensor on every iteration - after the first warmup pass that tensor sits in L2, so the bench rewards configs that exploit cache reuse (wider thread layouts, deeper smem stages, TMA descriptors) which lose under production cache-cold workloads. Replaced with `_l2_cold_cuda_graph_bench`:

- **Warmup**: plain Python loop running the kernel N times round-robin over the pre-cloned input sets -- drains P-state transitions, warms launch path / kernel compile / smem alloc.
- **Timed window**: single `graph.replay()` of a captured CUDA graph whose body records N round-robin kernel launches -- no Python loop, no per-launch CPU overhead. `ms/call = elapsed / total_recorded_launches = aggregate bytes / aggregate replay time / device peak`.
- Cloned tensor sets allocated once per shape in `__call__.benchmark()` and shared across all ~400 configs (avoids 400x re-cloning per shape).
- Falls back to legacy `triton.do_bench` path when a custom `do_bench=` is supplied or pre/post hooks are configured (those hooks' clone/restore doesn't survive CUDA graph capture).

Without this fix the autotuner regularly mis-ranks at heuristic boundaries (e.g. `row_bytes = 16K`): an earlier run showed 16 cells regressing 5-10% across 256 shapes. With the fix, that drops to 2.

## Benchmarks on GB200

Measured on a single GB200 (NVIDIA HBM3e, 1020W) across **256 shapes** - `T x D in {1K, 2K, 4K, 8K, 12K, 16K, 24K, 32K}` x 4 dtype combos `{bf16/bf16, fp32/fp32, bf16/fp32, fp32/bf16}`. Baseline = current `quack/main` (analytical heuristic, post-#132). Comparison = `rmsnorm_bwd_tuned` with the autotune cache warmed once per shape.

### Speedup heatmap (4 dtype combos x TxD)

Fwd:

image

- **Geomean: +6.7%** 

Bwd:

image

- **Geomean: +4.9%**
- **Median: +2.3%**
- **Max: +31.9%**
- **Min: -5.8%**
- **Overall: 148 / 256 shapes faster** (>=0.5%), **0 shapes regressing > 10%**, **2 shapes regressing 5-10%** (`bf16/fp32 T=24K D=8K` and `bf16/bf16 T=32K D=12K`)

Per-dtype geomean: bf16/bf16 **+7.4%**, fp32/fp32 **+5.0%**, bf16/fp32 **+3.9%**, fp32/bf16 **+3.4%**.

Top wins concentrate in the `D in {2K, 4K}` band (+23 to +32% across bf16/bf16 and fp32/fp32) - exactly the corners #132's TODO list called out as conservative. The autotuner finds non-obvious config combinations the analytical ladder misses; biggest single win is +31.9% at `T=4K, D=4K, bf16/bf16`.

### %SOL utilization - main vs this MR

Same 256-shape grid expressed as **% of GB200 HBM3e peak (8 TB/s per GPU)**, on a fixed 0-100 RdYlGn scale so the two plots are directly cell-by-cell comparable.

main (analytical heuristic, post-#132):

image

This MR (autotuner):

Bwd:

image

| dtype combo | main geomean %SOL | main peak %SOL | tuned geomean %SOL | tuned peak %SOL |
|-------------|-------------------|----------------|--------------------|-----------------|
| bf16/bf16   | 52                | 87             | 56                 | 86              |
| fp32/fp32   | 77                | 91             | 81                 | 91              |
| bf16/fp32   | 73                | 91             | 76                 | 90              |
| fp32/bf16   | 76                | 89             | 79                 | 89              |

Peak %SOL is essentially unchanged (the analytical heuristic already saturates bandwidth at the easy corners). Geomean lifts **+3-4 pp across all four dtype combos** - gains are concentrated in the mid-band where the autotuner finds non-obvious config combinations the analytical ladder misses.

## Tests

- 18 fwd + 16 bwd config-direct tests parametrize specific launch combinations (`reload_from`, `cluster_n`, `(num_threads, threads_per_row)` for fwd; same plus `reload_x`, `reload_wdy` for bwd) and verify numerical equivalence against `rmsnorm_ref` / `rmsnorm_bwd_ref` with bf16 tolerance x 2 (different configs pick different reduction trees).
- 2 end-to-end dispatch tests invoke `rmsnorm_{fwd,bwd}_tuned` without `.fn`, exercising the full `@autotune` flow (early-prune -> bench -> cache -> call) on M=N=256.
- 2 guard tests verify `.fn` raises `RuntimeError("...requires a config")` when called bare (without the decorator's config injection).

### DIFF
--- a/quack/autotuner.py
+++ b/quack/autotuner.py
@@ -81,24 +81,30 @@ def _l2_cold_cuda_graph_bench(
     arg_sets,
     kwarg_sets,
     extra_kwargs,
-    n_warmup_calls: int = 50,
+    warmup_target_ms: float = 200.0,
     n_timed_calls: int = 200,
     quantiles=None,
 ):
     """L2-cold single-replay CUDA-graph benchmark.
 
-    Warmup runs ``n_warmup_calls`` kernel invocations round-robin over the
-    pre-cloned ``(arg_sets[i], kwarg_sets[i])`` pairs as a plain Python loop
-    (drains P-state transitions and warms the launch path / kernel compile /
-    smem alloc). The timed window is a single ``graph.replay()`` of a
-    captured CUDA graph whose body records ``n_timed_calls`` round-robin
-    invocations — no Python loop, no per-launch CPU overhead — so the
-    measurement is just GPU work / total recorded calls.
+    Warmup is time-based: probe a single kernel launch to estimate the
+    per-call cost, then iterate round-robin over the pre-cloned
+    ``(arg_sets[i], kwarg_sets[i])`` pairs as a plain Python loop for
+    ``warmup_target_ms`` of wall-clock GPU work. Heavy pipelined configs
+    (TMA + smem_stages=3) need enough warmup to drain the pipeline-fill
+    phase or the timed window catches them artificially fast - a fixed
+    count of warmup launches underwarms heavy configs while overpaying
+    for cheap ones.
 
-    Round-robin over fresh tensor sets defeats the L2-resident caching that
-    inflates short-kernel timing under ``triton.testing.do_bench`` (which
-    calls the kernel on the same single tensor each iteration). For
-    cache-cold production workloads the round-robin number predicts
+    The timed window is a single ``graph.replay()`` of a captured CUDA
+    graph whose body records ``n_timed_calls`` round-robin invocations -
+    no Python loop, no per-launch CPU overhead - so the measurement is
+    just GPU work / total recorded calls.
+
+    Round-robin over fresh tensor sets defeats the L2-resident caching
+    that inflates short-kernel timing under ``triton.testing.do_bench``
+    (which calls the kernel on the same single tensor each iteration).
+    For cache-cold production workloads the round-robin number predicts
     real-world latency; the L2-hot number favours wider layouts / deeper
     smem stages that don't actually win once data has to come from HBM.
 
@@ -114,6 +120,23 @@ def _l2_cold_cuda_graph_bench(
     # Round timed-call count to a multiple of n_sets for even L2 turnover.
     rounds_timed = max(1, n_timed_calls // n_sets)
     total_timed_calls = rounds_timed * n_sets
+
+    # A few priming launches so the probe doesn't catch first-launch
+    # driver / kernel-load overhead.
+    for _ in range(3):
+        fn(*arg_sets[0], **kwarg_sets[0], **extra_kwargs)
+    torch.cuda.synchronize()
+
+    # Probe a single launch to estimate per-call ms; size the warmup loop
+    # to hit ``warmup_target_ms`` of GPU work regardless of kernel cost.
+    probe_start = torch.cuda.Event(enable_timing=True)
+    probe_end = torch.cuda.Event(enable_timing=True)
+    probe_start.record()
+    fn(*arg_sets[0], **kwarg_sets[0], **extra_kwargs)
+    probe_end.record()
+    torch.cuda.synchronize()
+    est_ms = max(probe_start.elapsed_time(probe_end), 1e-3)
+    n_warmup_calls = max(50, int(warmup_target_ms / est_ms))
 
     # Warmup: plain Python loop over rotating sets, no graph capture.
     for i in range(n_warmup_calls):
--- a/quack/autotuner.py
+++ b/quack/autotuner.py
@@ -13,6 +13,11 @@ import json
 from pathlib import Path
 from functools import cached_property, partial
 from typing import Dict, Tuple, List, Optional, Any
+from quack.bench.bench_utils import (
+    _bench_cuda_graph_l2_rotate,
+    _clone_l2_rotate_inputs,
+    _pick_l2_rotate_count,
+)
 
 import torch
 from torch import Tensor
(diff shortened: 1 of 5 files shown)


quack/autotuner.py
```diff
@@ -75,6 +76,146 @@ def _base32(key):
     return base64.b32encode(bytes.fromhex(key)).decode("utf-8").rstrip("=")
 
 
+def _l2_cold_cuda_graph_bench(
```
tridao: this is useful so it should prob be in quack/bench/

quack/rmsnorm.py
```diff
@@ -53,18 +62,31 @@ def _ensure_contiguous(t):
...
         self.is_layernorm = is_layernorm
-        self.reload_from = None if N <= (16384 if is_layernorm else 8192) else "smem"
-        self.delay_w_load = False
+        if config is None:
+            config = RmsNormFwdConfig.from_analytical_heuristic(
+                N, dtype.width, is_layernorm=is_layernorm
+            )
+        self.config = config
+        self.reload_from = config.reload_from
+        self.delay_w_load = config.delay_w_load
```
tridao, line 80: do we even want to tune this?

quack/rmsnorm.py
```diff
@@ -24,7 +24,16 @@
 from quack.reduction_base import ReductionBase
 from quack.cache_utils import jit_cache
```
tridao: latest main has `from quack.cache import jit_cache`

quack/autotuner.py
```diff
@@ -75,6 +76,146 @@ def _base32(key):
...
+    for _ in range(n_graphs):
+        arg_sets.append(tuple(a.clone() if isinstance(a, Tensor) else a for a in args))
+        kwarg_sets.append({
+            k: v.clone() if isinstance(v, Tensor) else v for k, v in kwargs.items()
+        })
+    return arg_sets, kwarg_sets
+
+
+def _pick_n_graphs_for_l2_cold(
+    args, kwargs, target_ratio: int = 3, min_graphs: int = 4, max_graphs: int = 16
```
tridao: are these default values broadly applicable, or do thye need to changed for different functions (e.g. gemm vs rmsnorm)?

tridao (approved):

tridao:
Really well-done, thanks @thakkarV ! You can just rebase and merge.

quack/rmsnorm.py
thakkarV: probably not, can remove. I did not find a huge gain from this so it almost certainly does not justify 2xing the tuning cost.

quack/autotuner.py
thakkarV: this I have not totally made sure of, but I think they should be yes. They are more a function of the tensor footprints than algorithm specific.

quack/autotuner.py
thakkarV: done.

PR #139, opened 2026-05-15
Remove deprecation warning in 4.5

drisspg:
See title, removing

```Shell
/home/drisspg/.venvs/flex/lib/python3.13/site-packages/nvidia_cutlass_dsl/python_packages/cutlass/cute/core.py:5789: DeprecationWarning: Use explicit `struct.scalar.ptr` for pointer instead.
  warnings.warn
  ```


tridao:
This would need to wait for after we update to 4.5.x since it's not compatible with 4.4.x

tridao (approved):

PR #143, opened 2026-05-19
Add non-gated colvec_scale support to gemm_dact

JungHoyoun:
I would like to add a `colvec_scale` argument to the `gemm_dact` signature, similar to `gemm_dgated`.

The motivation is to add `squared_relu` activation support to `sonic-moe`. Currently, `sonic-moe` only supports GLU-family activations, and this change is intended to extend support to architectures such as Nemotron 3.

The initial commit was written with the help of Codex. I will clean up and verify the code properly before moving this PR out of draft.

So far, testing has only been run on an SM90 environment.

Also, #142 was accidentally pushed while I was working with an AI tool. I have closed that PR, but I would appreciate it if it could be deleted from the repository history if possible. Sorry for the noise.


JungHoyoun:
This change adds `colvec` support to `gemm_dact` for direct-activation backward epilogues. This is useful when the backward path needs a per-row/per-token scale and optional reduction, for example MoE MLPs with `squared_relu` in model families like Nemotron.

The key behavior is:

- The implementation follows the existing `gemm_dgated` colvec path where the behavior overlaps.
- `colvec_scale` scales both `dx`/`dh` and the recomputed `postact`.
- `colvec_reduce` accumulates `sum(raw * act(h))`, using the unscaled GEMM accumulator as needed for router-score gradients.
- The no-colvec path preserves the existing default epilogue behavior.
- The added tests compare against PyTorch float32 references for both regular and varlen-M `gemm_dact`.

For validation, I checked that the without-colvec path stays aligned with clean HEAD, including identical generated SASS and matching NCU profiles. This was tested on SM90.

The relevant correctness tests are:

```bash
PYTHONPATH=$PWD pytest \
  tests/test_linear.py::test_gemm_dact \
  tests/test_linear_varlen_m.py::test_gemm_dact_varlen_m \
  -x -s
```

스크린샷 2026-05-31 오후 10 32 51


There is some duplicated structure between `gemm_dact` and `gemm_dgated` now, so a small refactor may be possible later, but I kept that out of this PR to keep the functional change focused.

tridao:
can you fix the lint? Then I can get CI gpu test to run

JungHoyoun:
Thanks, I cleaned up the lint.

JungHoyoun:
Hi @tridao, Could you please run the CI GPU tests?

tridao:
This is now implemented on main.

PR #144, opened 2026-05-24
[Rotary] Fuse copy into kernel

simveit:
Remove blocking host side computation by putting it into the kernel and add a benchmark for rotary.

Following copy is quiet expensive, this PR removes it

```python
    out = x if inplace else torch.empty_like(x)
    if rotary_dim < headdim and not inplace:
        out[..., rotary_dim:].copy_(x[..., rotary_dim:])
```

Benchmark:

| Path | Shape `(B,S,H,D,rotary_dim)` | Main ms | PR ms | Speedup | Main GB/s | PR GB/s | GB/s Delta |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| fwd | `(8,4096,32,128,128)` | 0.0921 | 0.0921 | 1.00x | 5841 | 5840 | -0.0% |
| fwd | `(8,4096,32,128,64)` | 0.1931 | 0.0999 | 1.93x | 2783 | 5382 | +93.4% |
| fwd | `(4,4096,32,256,128)` | 0.1910 | 0.0963 | 1.98x | 2816 | 5588 | +98.4% |
| fwd | `(2,4096,32,512,256)` | 0.1909 | 0.0938 | 2.04x | 2823 | 5747 | +103.6% |
| bwd | `(8,4096,32,128,128)` | 0.0921 | 0.0921 | 1.00x | 5839 | 5838 | -0.0% |
| bwd | `(8,4096,32,128,64)` | 0.1930 | 0.0997 | 1.94x | 2785 | 5388 | +93.5% |
| bwd | `(4,4096,32,256,128)` | 0.1937 | 0.0982 | 1.97x | 2777 | 5476 | +97.2% |
| bwd | `(2,4096,32,512,256)` | 0.1912 | 0.0933 | 2.05x | 2818 | 5777 | +105.0% |

### DIFF
--- a/benchmarks/benchmark_rotary.py
+++ b/benchmarks/benchmark_rotary.py
@@ -1,0 +1,406 @@
+import argparse
+import math
+import os
+
+os.environ.setdefault("TORCH_COMPILE_DYNAMIC", "0")
+
+import torch
+from triton.testing import Benchmark, do_bench, perf_report
+
+from quack.bench.bench_utils import run_and_print
+from quack.rotary import apply_rotary_emb
+
+
+# Keep the flattened row count (B * S * H)
+# fixed while increasing D, then shrink rows for large D to keep the
+# benchmark tensor size bounded.  B/S/H stay explicit in the table so
+# sequence/head-layout effects remain visible.
+SHAPE_CASES = {
+    "rows1m-d32-full": (8, 4096, 32, 32, 32),
+    "rows1m-d64-full": (8, 4096, 32, 64, 64),
+    "rows1m-d96-full": (8, 4096, 32, 96, 96),
+    "rows1m-d128-full": (8, 4096, 32, 128, 128),
+    "rows1m-d128-half": (8, 4096, 32, 128, 64),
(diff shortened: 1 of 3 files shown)


tridao:
if (D, rotary_dim) = (128, 64), then doesn't the kernel in main only read 64 elements and write 64 elements, and copy reads 64 elements and writes 64 elements? So in terms of IO, the version in main is fine?
I'm just trying to understand where the suboptimality comes from.

simveit:
my naive thought why that gives such a good boost in performance is that the op itself is very cheap and thus the overhead of launching a copy and the additional ordering within the whole stream we run on may dominate. i could check this claim tomorrow in nsight system.

simveit:
Screenshot 2026-05-25 at 19 55 42
Screenshot 2026-05-25 at 19 56 16
Here I attach two images from nsight systems (once for this PR state and once the old rotary). The launch between to cutlass kernels becomes significantly larger because we also launch the elementwise in between. There is also some significant idle time between the launch of the kernel and elementwise kernel in current main

PR #148, opened 2026-06-04
add mxfp8 gemm gated and dgated in sm100

WonsukYangTrlbs:
This PR adds `GemmGatedBlockscaledSm100` and `GemmDGatedBlockscaledSm100` for MXFP8 version of `gemm_gated` and `gemm_dgated`. Tested on B200 only.


tridao:
This is now implemented on main.

PR #149, opened 2026-06-10
[Autotuner] Fix BrokenPipeError root cause

akshattripathi-art:
Hi, I noticed that some gemm configs weren't compiled due to caught `BrokenPipeError`s from https://github.com/Dao-AILab/quack/blob/99bd7973bf3dc6db40961e413d4bdfea6c6fee3e/quack/autotuner.py#L446-L459

This turned out to happen when we sent tensors in the `kwargs`, which would crash when other non 0 ranks would try to load them up. Switching to using `tensor_meta` like `args` ended up fixing this.

### DIFF
--- a/quack/_compile_worker.py
+++ b/quack/_compile_worker.py
@@ -167,8 +167,14 @@ def main():
                     fake_args.append(_make_fake_tensor(meta))
                 else:
                     fake_args.append(meta)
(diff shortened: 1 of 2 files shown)


metalcycling (approved):
LGTM

PR #151, opened 2026-06-12
Fix compatibility issues with CuTe DSL 4.6.0+

anakinxc:


tridao:
Thanks! I guess we still have to wait for the PyPI wheel for 4.6.0.dev0?

anakinxc:
> Thanks! I guess we still have to wait for the PyPI wheel for 4.6.0.dev0?

Yes

anakinxc:
wheel is out, I think we can run ci now.

tridao (approved):

PR #152, opened 2026-06-12
Epi Mixins: Add support for variadic aux outputs for epi visitors

thakkarV:
Generalize the epilogue visitor API from a single optional aux output to a tuple of aux outputs so one GEMM can emit several epilogue tensors

- epi_setup_aux_out returns a tuple of (tiled_copy_r2s, tRS_sAuxOut, copy_aux_out) triples; () means no aux output.
- epi_visit_subtile returns a tuple of register tensors, one per aux output, matching the setup tuple ((), instead of None, by default).
- epi_convert_aux_out takes the constexpr output index as its first argument so multi-output mixins can pick per-output conversions.

The epilogue loops in GemmBase and GemmSymmetric convert and store each aux output in turn; all outputs share the epi_buffer index so the s2g TMA stores happen in lockstep after the smem fence. Existing single-output mixins (Act/Gated/DAct/DGated/NormAct/NormGated/SqReduce) now return one-element tuples; behavior is unchanged.

### DIFF
--- a/quack/activation.py
+++ b/quack/activation.py
@@ -21,17 +21,24 @@ sub_packed_f32x2 = partial(
 
 
 @dsl_user_op
-def tanh(a: float | Float32, *, loc=None, ip=None) -> Float32:
-    return Float32(
-        llvm.inline_asm(
-            T.f32(),
-            [Float32(a).ir_value(loc=loc, ip=ip)],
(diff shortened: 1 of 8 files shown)


tridao:
Really well done, thank you!

tridao (approved):

PR #145, opened 2026-05-25
[SM120] Add NVFP4 blockscaled GEMM path (~95% CUDA)

alecco:
**This PR requires https://github.com/NVIDIA/cutlass/pull/3273.**

CUDA example "79a_blackwell_geforce_nvfp4_bf16_gemm" ~289 TFLOP/s.

```
$ SM120_NVFP4_PATH=fast WARMUP=1 ITERS=1 ./scripts/run_sm120_nvfp4_bench.sh
279.6 TFLOP/s
```

## Summary

This PR adds a narrow SM120 NVFP4 blockscaled GEMM path to QuACK.

It follows the existing blockscaled compile/test facade where possible, but the kernel implementation is SM120-specific. SM120 does not use the SM100 `tcgen05` / TMEM path, so this PR builds the SM120-native path around:

- packed K-major `Float4E2M1FN` A/B storage
- compact 1D interleaved `Float8E4M3FN` scale storage
- native A/B TMA
- native FP8 scale TMA
- bundled MXF4/NVFP4 warp MMA
- split ping-pong TMA producer paths
- direct global BF16 stores in the default validated path

This PR depends on the corresponding CUTLASS CuTe DSL SM120 MXF4/NVFP4 native-TMA support.

## Supported configuration

The public SM120 blockscaled path is intentionally narrow:

- device: SM120
- A/B dtype: `Float4E2M1FN`
- scale dtype: `Float8E4M3FN`
- scale vector size: `16`
- output dtype: `BFloat16`
- A/B layout: K-major packed FP4
- D layout: N-major BF16
- scale storage: compact 1D interleaved FP8
- tile shape: `128x128x128`
- cluster shape: `1x1`
- no varlen M/K

Unsupported SM120 blockscaled configurations fail early instead of falling through to the SM100 path.

## Host storage and facade

Adds SM120 NVFP4 host-side helpers for:

- validating packed FP4 A/B tensors
- validating compact interleaved FP8 scale storage
- validating BF16 N-major output tensors
- creating packed FP4 A/B test tensors
- creating compact interleaved scale tensors
- creating TensorFill-like nonzero FP4/scale test inputs

The compile path routes the supported SM120 NVFP4 configuration through `compile_blockscaled_gemm_tvm_ffi(...)` with early validation of A, B, D, SFA, and SFB.

The public scale validator intentionally rejects the older rank-4 physical scale tensor form, so callers cannot pass storage that the kernel would reinterpret as compact interleaved scale storage.

## Kernel path

[...]

PR #153, opened 2026-06-15
[NVIDIA] Add SM120 (Blackwell GeForce) block-scaled GEMM (MXFP8 + FP4/NVFP4)

johnnynunez:
## Summary

Brings block-scaled GEMM to **SM120** (Blackwell GeForce / RTX PRO 6000 / RTX 50),
reaching parity with the SM100 path for the dense, K-major **MXFP8** case
(the torchao / `torch._scaled_mm` common case).

Before this PR, `compile_blockscaled_gemm_tvm_ffi` raised
`RuntimeError("Blockscaled SM100 GEMM requires SM100/SM110")` on SM120 — even
though SM120 hardware has the block-scaled warp MMA (`mma.sync.kind::mxf8f6f4`).

## Why a separate kernel

SM100 block-scaling uses tcgen05 + TMEM, which SM120 does **not** have, so the
SM100 code can't be shared. This PR vendors NVIDIA's CUTLASS DSL geforce
reference (`examples/python/CuTeDSL/cute/blackwell_geforce/kernel/blockscaled_gemm/`)
into QuACK and dispatches to it when device major == 12:

- `quack/gemm_sm120_blockscaled.py` — `Sm120BlockScaledGemmKernel`: warp-level
  block-scaled MMA + `ldmatrix`, TMA load of A/B/SF, 16-bit STMatrix epilogue.
  BSD-3 header preserved.
- `quack/sm120_blockscaled_dispatch.py` — MMA-op + ldmatrix-atom dispatch.
- `quack/blockscaled_gemm_utils.py` — `compile_blockscaled_gemm_tvm_ffi`
  dispatches to `_compile_blockscaled_gemm_sm120` on SM120, with a clear
  `NotImplementedError` for out-of-scope configs.

The SF packed layout (`(l, rm, rk, 512)` BlockScaledBasicChunk) is identical
between QuACK's packer, the SM100 path, and the SM120 reference, so scale
factors are consumed without repacking (strided slices are made contiguous).

## Scope / capability boundary

Supported on SM120: **dense, K-major MXFP8** (e4m3/e5m2 + e8m0 scales,
sf_vec=32) with f16/bf16 output, CTA tile 128×128×128.

Still SM100/SM110-only (raise `NotImplementedError`; tests skip on SM120):
FP4/NVFP4/MXFP4/mixed FP4×FP8, Float32 output (16-bit STMatrix epilogue),
non-K-major A/B (fp8 transpose ldmatrix unaligned on SM120), varlen.

## Testing

`tests/test_gemm_sm100_blockscaled.py` on RTX PRO 6000:
**34 passed, 23 skipped, 0 failed** (skips are the SM120 capability boundary
above). Numerical correctness vs `blockscaled_gemm_reference`, max_err ~1e-3
for bf16 output. SM100/SM110 behavior is unchanged.

## Perf (RTX PRO 6000, MXFP8 e4m3 → bf16)

| M | N | K | TFLOPS |
|---|---|---|--------|
| 2048 | 2048 | 2048 | 361 |
| 4096 | 4096 | 4096 | 497 |
| 8192 | 8192 | 8192 | 536 |

Peak ~536 TFLOPS @ 8192³ — above the SM120 bf16 tensor-core roofline
(~390 TFLOPS), confirming the FP8 ~2× win.


johnnynunez:
## Relation to #127 / #128, and updates in this PR

Flagging the overlap with @alecco's #127 (SM120 FP4 correctness) and #128 (SM120 FP4 packed-LDSM perf) so this can be triaged cleanly. All three are open; this PR was developed independently and differs in approach and scope.

### Approach
- **This PR**: vendors NVIDIA's standalone CuTe-DSL *geforce* block-scaled reference as a separate `Sm120BlockScaledGemmKernel` and dispatches to it from `compile_blockscaled_gemm_tvm_ffi` when `device major == 12`. Works on **stock `nvidia-cutlass-dsl[cu13]` 4.5.2**.
- **#127 / #128**: extend the existing `GemmSm120` class in-place and **require the unmerged `NVIDIA/cutlass#3185`**.

### Scope / capability
| | this PR (#153) | #127 | #128 |
|---|---|---|---|
| MXFP8 | ✅ | ❌ | ❌ |
| MXFP4 / NVFP4 | ✅ | ✅ | ✅ |
| Output dtypes | f16 / bf16 / **f32** | bf16 | bf16 |
| CUTLASS#3185 needed | **no** | yes | yes |
| Shapes | M/N any, tile 128×128×128 | M%128, N%128, K%64 | + 64×64 tiles |

### Two findings worth surfacing
1. **Float32 output** initially produced ~74% garbage and looked unsupported. Root cause: the epilogue built the C copy atom from `c_dtype`, but that atom only sets the register→smem *tiling geometry* (`make_tiled_copy_C_atom`) — the real store is `copy_atom_r2s`, which `sm120_get_smem_store_op` already picks correctly. Building the C atom from `Float16` unconditionally (as `gemm_sm90` does) fixes it; f32 output is now exact (rel_err 0.0). **The same latent bug exists in the NVIDIA geforce reference.**
2. **Compact-scale TMA trap** (the sub-16-byte row-stride issue #127 guards against with padded row-major scales) does **not** apply here: QuACK's `(l, rm, rk, 512)` BlockScaledBasicChunk SF layout always has a 512-byte contiguous inner tile. Verified across `sf_k ∈ {4,8,12,16,20,32,64,128}` (incl. non-mult-of-16) for all three formats — all pass.

### Tests (RTX PRO 6000, sm_120)
`tests/test_gemm_sm100_blockscaled.py`: **59 passed, 16 skipped, 0 failed**. Skips are genuine boundaries: 12 varlen, 3 non-K-major, 1 invalid FP4/sf-dtype combo. SM100/SM110 unchanged.

### Perf (RTX PRO 6000, → bf16)
| fmt | M=N=K | TFLOP/s |
|-----|-------|---------|
| mxfp8 | 8192 | 507 |
| mxfp4 | 8192 | 1431 |
| nvfp4 | 8192 | 1364 |

FP8 ~500-530, FP4 ~1400 TFLOP/s (above the sm120 bf16 roofline ~390). f32 output is exact at ~770-800 TFLOP/s for FP4. Full sweep + the compact-scale probe are attached in the branch under `profiling/sm120_mxfp8_blockscaled/`.

Happy to rebase, narrow scope, or coordinate with #127/#128 — whatever's easiest for review.

alecco:
> ## Relation to #127 / #128, and updates in this PR
> ...
> Happy to rebase, narrow scope, or coordinate with #127/#128 — whatever's easiest for review.

Nice! SM120 needs some love.

My PR on CUTLASS didn't get much attention, so I really don't mind. And I wouldn't be surprised if Nvidia is working on a different way to implement it so it's a bit of a gamble.

I strongly recommend checking the generated PTX/SASS as good TMA instructions are tricky to get right.

FWIW I compared my PR vs CUDA example 79a for reference (~95% TFLOPS).

johnnynunez:
> > ## Relation to #127 / #128, and updates in this PR
> > ...
> > Happy to rebase, narrow scope, or coordinate with #127/#128 — whatever's easiest for review.
> 
> Nice! SM120 needs some love.
> 
> My PR on CUTLASS didn't get much attention, so I really don't mind. And I wouldn't be surprised if Nvidia is working on a different way to implement it so it's a bit of a gamble.
> 
> I strongly recommend checking the generated PTX/SASS as good TMA instructions are tricky to get right.
> 
> FWIW I compared my PR vs CUDA example 79a for reference (~95% TFLOPS).

Yes, i am trying to full support sm120. The experts here are @tridao @drisspg @gau-nernst

i have an overview about cuda, cutlass etc and I used a lot nvidia nisght in the past, so i am not expert at all, but now with CuTe DSL and AI + agent era + hardware in the loop it is easier to integrate that things.

i have also 4 prs in flash attention, it is possible to get extra performance compared baseline implementations

gau-nernst:
Perf doesn't look very good I think. This is what I got using my kernel here https://github.com/gau-nernst/gn-kernels/blob/main/gn_kernels/cutedsl/sm120_mm_mxfp8.py on 5090 (PRO6000 should be around 10% faster)

MXFP8

M | N | K | torch._scaled_mm() | Mine
---|---|---|---|---
2048 | 2048 | 2048 | 552.61 | 483.69
4096 | 4096 | 4096 | 678.47 | 636.19
8192 | 8192 | 8192 | 642.64 | 652.53

NVFP4

M | N | K | torch._scaled_mm() | Mine
---|---|---|---|---
2048 | 2048 | 2048 | 956.36 | 796.73
4096 | 4096 | 4096 | 1310.21 | 1212.52
8192 | 8192 | 8192 | 1342.81 | 1341.16

johnnynunez:
> Perf doesn't look very good I think. This is what I got using my kernel here https://github.com/gau-nernst/gn-kernels/blob/main/gn_kernels/cutedsl/sm120_mm_mxfp8.py on 5090 (PRO6000 should be around 10% faster)
> 
> MXFP8
> 
> M	N	K	torch._scaled_mm()	Mine
> 2048	2048	2048	552.61	483.69
> 4096	4096	4096	678.47	636.19
> 8192	8192	8192	642.64	652.53
> NVFP4
> 
> M	N	K	torch._scaled_mm()	Mine
> 2048	2048	2048	956.36	796.73
> 4096	4096	4096	1310.21	1212.52
> 8192	8192	8192	1342.81	1341.16

thanks as always! I will take a look!

johnnynunez:
> **[Correction — see follow-up comment below.]** The mechanism proposed in this comment is wrong. I measured it: the gap is **not** occupancy-capped by the 32KB smem epilogue, and a direct epilogue does **not** free smem to deepen the pipeline — `ab_stage` stays pinned at 2, occupancy stays 1 CTA/SM, and the direct epilogue buys only ~2% (after fixing store coalescing). The real gap is the persistent 9-warp mainloop structure vs gn-kernels' lean 5-warp per-CTA design. Full measured write-up in the follow-up comment. Original text kept below for the record.

---

Thanks @gau-nernst — I reproduced your finding on RTX PRO 6000 and dug into the SASS as you and @alecco both suggested. You're right, MXFP8 is the weak spot. Numbers below are this kernel vs `torch._scaled_mm` (cuBLAS), bf16 out, bit-exact:

| M=N=K | this PR (TFLOP/s) | cuBLAS (TFLOP/s) | ratio |
|-------|------|--------|-------|
| 2048 | 244 | 409 | 0.60 |
| 4096 | 473 | 629 | 0.75 |
| 8192 | 516 | 630 | 0.82 |

(FP4/NVFP4 is at parity with both your kernel and cuBLAS — `nvfp4@8192` ≈ 1364 here vs your 1341 vs cuBLAS 1342 — so MXFP8 is the gap to close.)

**SASS is clean** — 128 QMMA, 4 `UTMALDG`, 48 LDSM, no redundant `cp.async.bulk.tensor` from the TMA-folding trap. So it's not a bad-TMA-emission bug. The gap is structural, and the size-dependent shape pins it down:

1. **Wave quantization.** Persistent grid = 188 SMs, forced 128×128 tile → 2048 is only 1.36 waves (≈47% of SMs idle on the 2nd wave → 0.60x), while 8192 is 21.8 waves (quantization negligible → 0.82x).
2. **Occupancy hard-capped at 1 CTA/SM by smem.** `ab_stage=2` (64KB A/B) + a **32KB smem-staged TMA-store epilogue** + SF + mbars ≈ 97KB of the 100KB/SM budget. That smem epilogue is what blocks a deeper mainloop pipeline. Your kernel sidesteps this with a direct register→gmem `st.global` epilogue and 5 warps (vs the vendored reference's 9).

**Roadmap I'm taking:** rewrite the epilogue to direct register→gmem (drop the 32KB smem stage), which frees enough smem to push `ab_stage` 2→3 for better latency hiding at the large sizes where wave quantization is already gone. The small-size deficit is mostly inherent to the single forced 128×128 tile (cuBLAS switches tile shapes / split-K), so I'll likely document that boundary rather than chase it in this PR. Will report measured before/after.

Will ping you with the post-rewrite numbers — thanks again for the reference kernel and the SASS nudge, both were exactly right.

johnnynunez:
@gau-nernst — following up with the measured results, and I have to correct the mechanism I proposed in my previous comment.

Setup: RTX PRO 6000 Blackwell, torch 2.12.0+cu130, nvidia-cutlass-dsl 4.5.2, bf16 out, MXFP8 e4m3/e8m0. All paths bit-exact vs `blockscaled_gemm_reference` (max_err 5e-4–1e-3). Perf = `nsys cuda_gpu_kern_sum` (GPU kernel time, dispatch-noise-free), median of 100 iters.

Head-to-head (TFLOP/s, higher = better):

| M=N=K | this PR | gn-kernels | cuBLAS |
|-------|--------:|-----------:|-------:|
| 2048  | 366     | 490        | 551    |
| 4096  | ~495    | 602        | 609    |
| 8192  | ~533    | 637        | 619    |

I took the two levers from my earlier comment and actually measured them. Both turned out to be dead ends, so the earlier "occupancy-capped by the smem epilogue → free it to deepen the pipeline" story is **wrong** — here's the corrected version.

**1. Deeper pipeline (`ab_stage` 2→3) — not feasible.** A+B in FP8 is 32KB/stage, so 3 stages = 96KB ≈ the entire 100KB/SM budget, with nothing left for SF / mbars / epilogue. `_compute_stages` already correctly pins `ab_stage=2`. Smaller K-tiles don't rescue it either: `BK=64` won't compile (the block-scale SF atom needs `K%128==0` — `sf_vec=32 × 4-pack`), and `BK=256` gives `ab_stage=0`. So pipeline depth was never the lever — and your kernel runs `num_stages=2` too and still wins.

**2. Direct register→gmem epilogue — real but marginal (~2%), and only after fixing coalescing.** SASS, three variants at the 2048 tile:

| variant      | store SASS                       | result |
|--------------|----------------------------------|--------|
| staged (TMA) | 16 STSM + 16 STS + 2 STG         | baseline |
| direct naive | **64 STG.E.U16** (scalar 2-byte) | regresses (uncoalesced) |
| direct vec32 | **32 STG.E** (32-bit, 2×bf16)    | small win |

The naive direct store scattered each thread's accumulator into 64 separate 16-bit global writes (the 4×2×1 MMA atom over 8 warps gives non-contiguous per-thread gmem addresses). Forcing a 32-bit copy atom + a static C layout + `assumed_align=16` lets ptxas fuse N-adjacent bf16 pairs into 32 coalesced stores. Final nsys: **+0.6% @4096, +1.9% @8192**, slightly slower @2048. Coalescing was needed just to break even.

**The correction:** the gap is **not** occupancy-capped by the 32KB smem epilogue, and freeing that smem does **not** deepen the pipeline (`ab_stage` stays 2, occupancy stays 1 CTA/SM). The staged StMatrix+TMA epilogue was already efficient. The real difference vs your kernel is per-CTA structure: this vendored reference uses **9 warps (288 threads) + a 32KB smem-staged TMA-store epilogue** that pins occupancy at 1 CTA/SM, where yours uses **5 warps (4 MMA + 1 TMA) + a direct register→gmem epilogue** on a non-persistent 1-tile-per-CTA grid. The lean warp count + scheduling, not the store path, is the next thing to rework — and I'll measure it the same way before claiming anything.

(One cleanup for the record: an earlier "compute-only ≈ 919 TF" figure is **invalid** — SASS showed ptxas dead-code-eliminated the whole mainloop once the store output was unused, so it measured an empty kernel. Discarded.)

Thanks again for the reference kernel and the SASS nudge — both were right; I just attributed the gap to the wrong place.

alecco:
BTW I have newer draft PRs, with the same limitations and also those didn't get any attention, so also I don't mind either way.

[SM120] Add NVFP4 blockscaled GEMM path (~95% CUDA)
https://github.com/Dao-AILab/quack/pull/145

[CuTeDSL] Add SM120 MXF4/NVFP4 native-TMA path https://github.com/NVIDIA/cutlass/pull/3273

johnnynunez:
> BTW I have newer draft PRs, with the same limitations and also those didn't get any attention, so also I don't mind either way.
> 
> [SM120] Add NVFP4 blockscaled GEMM path (~95% CUDA) #145
> 
> [CuTeDSL] Add SM120 MXF4/NVFP4 native-TMA path [NVIDIA/cutlass#3273](https://github.com/NVIDIA/cutlass/pull/3273)

I think that quack was created first for sm100, and maybe runs better with that structure. I need @tridao here and his insights to take decisions and if I can ask internally this behavior

johnnynunez:
Following up with measured data — and correcting the mechanism in my earlier comment above (the "free smem → push `ab_stage` 2→3 / deepen pipeline" roadmap was wrong).

I built the direct register→gmem epilogue @gau-nernst suggested and measured it properly (nsys GPU kernel time, median of 100 iters, RTX PRO 6000, bf16 out, bit-exact). Two things turned out differently than my earlier comment claimed:

**1. The epilogue is not where the gap lives.** Freeing the 32KB smem-staged epilogue does *not* change `ab_stage` (stays pinned at 2) or occupancy (stays 1 CTA/SM), so "deepen the pipeline" was a dead end. The store is only ~4% of runtime by bandwidth.

**2. A naive direct store *regresses* — it has to coalesce first.** Dropping the StMatrix+TMA epilogue for a plain `st.global` emitted **64× `STG.E.U16`** (scattered 2-byte stores) — SASS-confirmed — because the 4×2×1 MMA atom over 8 warps scatters each thread's accumulator across non-contiguous gmem. That ran *slower* than the staged path.

Forcing coalescing (raw C tensor with a real iterator + a 32-bit `CopyUniversalOp` atom fusing the N-adjacent bf16 pair + static C layout with `assumed_align=16`) drops it to **32× `STG.E`** (32-bit). Then it's a small but real win:

| M=N=K | staged StMatrix+TMA | direct coalesced | gain |
|-------|--------------------:|-----------------:|-----:|
| 4096  | 277,869 ns          | 275,420 ns       | +0.9% |
| 8192  | 2,110,010 ns        | 2,017,080 ns     | +4.6% |

(tied at 2048). The coalesced direct store has no OOB predication, so I gate it automatically — direct when M,N % 128 == 0, staged fallback otherwise (a non-aligned shape like 256×192 produces wrong output on the direct path).

**Net:** the epilogue is a ~1–5% lever, not the bulk of the MXFP8 gap to your kernel (533 vs 637 TFLOP/s @ 8192). That gap is in the mainloop / kernel structure — this PR's persistent 9-warp scheme vs your non-persistent 1-tile-per-CTA + 5 warps (4 MMA + 1 TMA). That's where I'm looking next. Thanks again @gau-nernst for the reference kernel and the SASS nudge — both were exactly right, and the reference made this measurable.

johnnynunez:
Followed the gap into the mainloop as promised. Apples-to-apples now (both measured with nsys GPU kernel time on RTX PRO 6000, median of 100, same quantized inputs):

| M=N=K | this PR (direct epi) | @gau-nernst gn-kernels |
|-------|---------------------:|-----------------------:|
| 4096  | 498 TF               | 675 TF                 |
| 8192  | 531 TF               | 704 TF                 |

Both kernels do 128 QMMA/CTA, 1 CTA/SM, 2 stages — so it's per-CTA *execution efficiency*, not FLOP count. I ruled out the obvious structural levers with exact accounting before touching code:

- **Deeper pipeline (`ab_stage` 2→3): physically impossible.** Even with the epilogue smem fully freed, 3 stages need `3×(32768 A/B + 1024 SF) + 48 mbar = 101424 B` vs the SM120 hard opt-in cap of 101376 B — over by 48 bytes. So `ab_stage=2` is correct, and freeing smem genuinely buys nothing here.
- **2 CTAs/SM: blocked by both smem and registers** (135KB > 101KB; 2×133×288 regs > 65536). Occupancy is locked at 1.
- **Fewer MMA warps (8→4, atom (4,2,1)→(2,2,1)): regresses −2.9%** (REG 133→200, same QMMA concentrated into fewer warps). The 8-warp tiling is already well tuned.

The real difference shows up in the mainloop SASS: **this kernel emits ~80 scalar `LDS.U8` scale-factor loads per CTA (~11.5% of the mainloop instruction stream); gn-kernels emits zero** — it loads SF wide (`LDS.64`) via an Int32 recast + lane remap. I can't just widen the SF copy atom because this kernel's per-thread SF fragment is non-contiguous in the NVIDIA-swizzled chunk layout (stride-8, verifier rejects vectorization). Closing it means reworking the SF smem layout + `partition_fragment_SFA/SFB` to gn's contiguous-per-thread scheme. That's the next piece I'm picking up — wanted to share the measured root cause so it's on record. Thanks again @gau-nernst — your kernel is a great reference target.

tridao:
Thanks for this PR! This feature (mxfp8 and mxfp4 / nvfp4 for sm120) is now implemented as of v0.6.2. Perf should be ~cuBLAS (or slightly better) on RTX 5090.

PR #154, opened 2026-06-15
[NVIDIA] SM120 block-scaled GEMM: varlen (grouped) + batched FP4

johnnynunez:
## Summary

**Stacked on #153** (SM120 block-scaled MXFP8 + FP4/NVFP4 dense). Please merge #153 first — until then this PR's diff includes #153's commits; afterwards it collapses to the incremental changes below.

Closes two of the remaining SM120 capability gaps from #153:

### 1. varlen_m / varlen_k (grouped GEMM)
The vendored geforce kernel is a dense persistent kernel with **no grouped / cu_seqlens scheduler**. Rather than rewrite the scheduler, varlen is implemented as **per-expert dense dispatch** in the runner — each expert is launched as its own dense GEMM. This is correct because:
- the dense kernel already handles arbitrary, non-128-aligned M and K (verified);
- one dense kernel is compiled once and reused across experts via dynamic layouts (verified correct, including experts larger than the seed shape);
- scale factors use the same dQaccum-padded `BlockScaledBasicChunk` `(1, rm, rk, 512)` layout as the SM100 path — each expert `i` occupies a contiguous run of `rm` tiles (varlen_m) or `rk` tiles (varlen_k) at tile offset `cu[i] // 128 + i`, sliced directly from the packed buffer.

Scope: MXFP8 (matches the SM100 varlen operand builders), f16/bf16/f32 output. Per-expert launch has fixed overhead, so small experts are launch-bound; large / many-expert cases reach useful throughput. A fused SM120 grouped scheduler is future work.

### 2. batched (l>1) FP4
`_create_fp4_operand_tensor` packed `float4_e2m1fn_x2` with **L innermost** for `l>1`, which is not K-major and tripped the SM120 K-major guard. Fixed to emit a K-major layout (L outermost, K stride 1) via a byte-level reorder that preserves the 2-FP4-per-byte K packing. `l=1` is returned unchanged (byte-identical). Batched MXFP4 / NVFP4 now run.

## Testing (RTX PRO 6000, sm_120)
`tests/test_gemm_sm100_blockscaled.py`: **77 passed, 4 skipped, 0 failed** (was 59/16 in #153). Remaining skips: 3 non-K-major + 1 invalid FP4/sf-dtype combo — genuine boundaries. SM100/SM110 behaviour unchanged (the FP4 builder change is `l=1`-identical and only reorders `l>1` memory).

## Perf (varlen, MXFP8 → bf16, per-expert dispatch)
| mode | experts | shape | TFLOP/s |
|------|---------|-------|---------|
| varlen_m | 8 | 8×1024, N=K=4096 | 337 |
| varlen_m | 4 | [256,768,512,1024], N=K=2048 | 115 |
| varlen_k | 4 | [256,768,512,1024], M=N=2048 | 91 |

Throughput grows with expert size/count (amortizing launch overhead). Artifacts under `profiling/sm120_blockscaled_varlen_batched/`.


tridao:
This feature is now implemented as of v0.6.2 (1 kernel instead of launching a series of kernels).

PR #155, opened 2026-06-15
[NVIDIA] SM120 block-scaled GEMM: mixed FP4xFP8

johnnynunez:
## Summary

**Stacked on #153 and #154.** Please merge those first; until then this PR's diff includes their commits, after which it collapses to the incremental changes below.

Enables **mixed-precision FP4×FP8** on SM120: one operand FP4 (e2m1), the other FP8 (e4m3 / e5m2), with e8m0 scales at sf_vec_size=32, f16/bf16/f32 output.

## How
The vendored geforce kernel **already supported** `a_dtype != b_dtype` — `_setup_attributes` sets `mixed_mode`, allocates the FP4 operand as Int8 in SMEM, selects `MmaMXF8F6F4Op` (atom (16,8,32)) and the mixed-mode ldmatrix path (`LdMatrix8x16x8bOp`, `unpack_bits=4`, `<<2` FP4 shift). The only gap was the public interface passing a single `ab_dtype`.

This PR adds an optional **`b_dtype`** argument to `compile_blockscaled_gemm_tvm_ffi` (defaults to `ab_dtype` → all existing same-dtype calls unchanged). On SM120 the dispatch validates the pair is {FP4, FP8} with e8m0/vec32 scales and threads `b_dtype` into the kernel's A/B compile tensors. On SM100/SM110, a mixed pair raises `NotImplementedError` (kept SM120-only).

## Scope
A,B ∈ {Float4E2M1FN} × {Float8E4M3FN, Float8E5M2} (either order), e8m0 scales, vec32, K-major, f16/bf16/f32 out. Same-width mixed (e4m3+e5m2) and FP6 pairs remain unsupported (not in `MXF8F6F4_SUPPORTED_PAIRS`).

## Testing (RTX PRO 6000, sm_120)
`test_blockscaled_mixed_fp4_fp8`: 4 pairs × 3 shapes × {bf16, f32} = **24 cases, all pass** (rel_err ~3e-3). Same-dtype paths unchanged.

## Perf (→ bf16)
| pair | M=N=K | TFLOP/s |
|------|-------|---------|
| A=FP4 B=FP8 | 8192 | 554 |
| A=FP8 B=FP4 | 8192 | 540 |

Mixed sits between pure FP8 (~507) and pure FP4 (~1400) — the FP8 operand dominates load/MMA cost. Artifacts under `profiling/sm120_blockscaled_mixed_fp4_fp8/`.

PR #156, opened 2026-06-15
[NVIDIA] SM120 block-scaled GEMM: M-major A for FP8

johnnynunez:
## Summary

**Stacked on #153, #154, #155.** Enables **M-major A operands for FP8** on SM120 (the row-major-activations case), at parity performance with K-major.

## Root cause of the previous non-K-major block
`make_ldmatrix_atom` always used `LdMatrix8x8x16bOp` (a 16-bit StMatrix-style atom). With `transpose=True` on 8-bit data, its source needs a 128-bit-aligned 16-bit lane pattern, which an M-major FP8 SMEM tile (8-bit elements) cannot satisfy → ICE:
```
'cute.copy' op src ptr alignment (64 bits) does not meet requirement (128 bits)
```
This was **not** a hardware limit — the correct primitive for transposing 8-bit data is `ldmatrix.m16n16.trans.b8` (`LdMatrix16x16x8bOp`, `num_matrices=2`).

## Fix
In `make_ldmatrix_atom`, when `transpose and operand_dtype.width == 8` (non-mixed), use `LdMatrix16x16x8bOp(transpose=True, num_matrices=2)`. M-major A FP8 then compiles and is bit-correct.

## Scope / structural boundary
The warp MMA is **m16n8k32**:
- A (M) fragment is **16-wide** → meets the 16 elements `ldmatrix.m16n16.trans.b8` needs → **M-major A works**.
- B (N) fragment is **8-wide** → below 16 → **N-major B cannot** feed the 8-bit transpose, stays K-major-only.
- FP4 (4-bit) has no ldmatrix transpose primitive → **FP4 A stays K-major**.

Net SM120 major support: **A K-major (any dtype) or M-major (FP8)**, **B K-major only**. This covers the common case (A = activations, possibly row-major; B = weights, storable K-major). N-major B would need loading 2 N-tiles together / a different fragment layout — a deeper mainloop change for little practical gain.

## Testing (RTX PRO 6000, sm_120)
`test_gemm_sm100_blockscaled.py`: **102 passed, 3 skipped, 0 failed** (was 77/4). `test_blockscaled_mxfp8_major_modes` now runs A M-major on SM120. Remaining skips: 2 N-major-B + 1 invalid FP4/sf combo.

## Perf (MXFP8 → bf16)
| A major | M=N=K | TFLOP/s |
|---------|-------|---------|
| K-major | 8192 | 543.9 |
| M-major | 8192 | 544.6 |

M-major A is at **parity** — the m16n16.trans.b8 atom adds no cost. Artifacts under `profiling/sm120_blockscaled_nonkmajor_fp8/`.


tridao:
This feature is now implemented as of v0.6.2. We also support B being k-major or n-major.

PR #157, opened 2026-06-16
Split k gemm sm100

santoshmo:
## Split-K GEMM for SM100 (B200/B300)

  Adds split-K to the CuTe-DSL GEMM: each output tile's K reduction is split across `split_k`
  work units, so K-heavy small-M/N shapes (which under-fill the GPU at one CTA per tile) spread
  work across many more CTAs. Two run-to-run-deterministic modes:

  - **`parallel`** (default): splits write fp32 partials to a workspace; a separate table-driven
  reduce kernel sums them and applies the epilogue (mirrors cuBLAS's `splitKreduce_kernel`).
  Scales with `split_k`.
  - **`serial`**: fused in-kernel turnstile reduction (single kernel).

  `split_k=0` auto-tunes per shape. The reduce is table-driven (Stream-K-ready) and uses an
  intra-CTA K-split to fill the GPU on few-tile reduces; the split-K `gemm()` is
  CUDA-graph-capturable.

  ### Performance (B200, bf16)

  Baseline is `torch.bmm` — which for these shapes is **cuBLAS's own split-K**
  (`nvjet_…_splitK_TNT` + `cublasLt::splitKreduce_kernel`), i.e. split-K vs split-K. `graph` =
  host-free (CUDA-graph replay) kernel comparison; `vs split_k=1` = speedup over a plain
  (non-split) GEMM.

  | shape (m×n×k) | vs cuBLAS (graph) | vs cuBLAS (eager) | vs `split_k=1` |
  |---|---|---|---|
  | 128×128×16384 | 0.98× | 0.95× | 2.90× |
  | 128×128×65536 | 1.05× | 0.90–0.98× | 7.88× |
  | 256×256×32768 | 1.09× | 1.01× | 3.84× |
  | 512×512×16384 | 0.89× | 0.85× | 2.05× |

  Host-free, quack matches/beats cuBLAS on 3 of 4 shapes; 512² is GEMM-bound (separate). Up to
  ~8× over a plain GEMM. (Shared, unlocked-clock B200 — ratios approximate; per-`split_k`
  speedups are stable.)

  ### Testing

  `tests/test_gemm_splitk.py` (32 cases): correctness vs fp32 ref (aligned / ragged-K / edge-tile
  / batched × both modes), zero-contribution splits, the reduce remainder + intra-CTA K-split
  paths, determinism, table-driven variable-contributor reduce, auto `split_k`, and CUDA-graph
  capture. Benchmark: `benchmarks/benchmark_gemm_splitk.py`.

### DIFF
--- a/quack/gemm.py
+++ b/quack/gemm.py
@@ -3,6 +3,7 @@
 
 from typing import Optional
 
+import torch
 from torch import Tensor
 
 import cutlass.cute as cute
@@ -11,7 +12,12 @@ from cutlass.cute.runtime import make_ptr
 
 from quack.cache_utils import jit_cache
 from quack.compile_utils import make_fake_tensor as fake_tensor
-from quack.cute_dsl_utils import get_device_capacity, get_max_active_clusters, torch2cute_dtype_map
+from quack.cute_dsl_utils import (
+    get_device_capacity,
+    get_max_active_clusters,
+    get_num_sms,
+    torch2cute_dtype_map,
+)
 from quack.gemm_default_epi import (
     GemmDefaultEpiMixin,
     GemmDefaultSm90,
@@ -19,6 +25,14 @@ from quack.gemm_default_epi import (
     GemmDefaultSm120,
 )
 from quack.rounding import RoundingMode
+from quack.gemm_splitk_reduce import (
+    auto_split_k,
+    choose_reduce_kgroups,
+    choose_reduce_vec_width,
+    compile_splitk_reduce,
+    splitk_reduce,
+    uniform_splitk_tables,
+)
 from quack.gemm_tvm_ffi_utils import (
     get_majors,
     get_dtypes,
@@ -63,6 +77,8 @@ def _compile_gemm(
     rounding_mode,
     sr_seed_mode,
     has_trace_ptr,
+    has_splitk=False,
+    splitk_parallel=False,
 ):
     sm_to_cls = {
         9: GemmDefaultSm90,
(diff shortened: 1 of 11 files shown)


quack/gemm.py
```diff
@@ -172,8 +186,13 @@ def gemm(
     sr_seed: int | Tensor = 0,
     use_tma_gather: bool = False,
     concat_layout: dict | None = None,
+    split_k: int = 1,  # K splits per output tile (SM100 only); 0 = auto (auto_split_k)
+    # "parallel": every split writes its own fp32 workspace slice, a second kernel
+    # reduces them (cuBLAS-style; split_k can scale to fill the GPU). "serial": fused
+    # in-kernel turnstile reduction (single kernel, but reduction cost grows with
+    # split_k). Both are run-to-run deterministic.
+    split_k_mode: str = "parallel",
```
thakkarV, line 194: this is split K serial still. parallel reduction is with atomics

quack/gemm.py
```diff
@@ -193,21 +214,50 @@ def gemm(
...
         assert tile_count_semaphore is not None, (
-            "Dynamic persistent tile scheduler for SM8x and SM90 requires a semaphore in GMEM"
+            "Dynamic persistent tile scheduler in SM90 requires a semaphore in GMEM"
         )
-    if device_capacity[0] == 8:
-        if add_to_output:
-            C = D
-            add_to_output = False
+
+    if split_k == 0:  # auto: pick a split_k for this shape (1 when split-K won't help)
```
thakkarV, line 227: zero should be considered invalid?

quack/gemm_sm90.py
```diff
@@ -1130,21 +1186,469 @@ def mma(
...
+                raster_order=scheduler_args.raster_order,
+                group_size=scheduler_args.max_swizzle_size,
+                tile_shape_mn=self.cta_tile_shape_mnk[:2],
+                cluster_shape_mnk=self.cluster_shape_mnk,
+                tile_count_semaphore=scheduler_args.tile_count_semaphore,
+                persistence_mode=persistence_mode,
+            )
+        return tile_sched_args
+
+    # ── Split-K with turnstile reduction ─────────────────────────────────────────
```
thakkarV, line 1430: please avoid the term "turnstile" - it is nonstandard (likely a claudeism). the standard term is serial reduction.

santoshmo: It was taken from https://research.colfax-intl.com/cutlass-tutorial-persistent-kernels-and-stream-k/. not sure if this written by Claude though

PR #158, opened 2026-06-16
[GEMM] Add SplitK Serial,Parallel,Separate modes to all GEMM classes and dtypes

thakkarV:
Adds split-K to the QuACK GEMMs (SM90/SM100/SM120) for small-M/N, large-K shapes that starve a 2D grid of occupancy. `split_k` and `split_k_mode` are exposed on every `gemm_interface.py` entry point and the autotuner surface (`GemmConfig.split_k`, `split_k=None` lets the tuner pick).

### Features

- `split_k` is a constexpr kernel specialization captured by the jit cache; `split_k == 1` compiles to exactly the pre-existing kernel (all new code is behind `const_expr`).
- The tile scheduler owns the K-mode decomposition: `gridDim.z` is scaled by `split_k` with the split index fastest-varying (splits of a tile stay L2-adjacent), balanced k-tile ranges, and a quack-owned `WorkTileInfo` that returns the decoded `(m, n, split, batch)` coordinate.
- The full epilogue (alpha, beta*C, bias, activations) runs exactly **once** per output tile, on the entity owning the final reduction determined by `is_splitk_leader`.
- Partials accumulate in f32, so the finalizer is dtype-agnostic: split-K works for bf16/fp16, tf32 (f32 inputs), and fp8 (e4m3/e5m2) inputs with any output dtype.
- Block-scaled MXFP8 GEMM composes with the same device path (SERIAL/PARALLEL) at no kernel cost - the scale-factor loads ride each split's K-range and the accumulator is already descaled f32 before the epilogue; only the block-scaled host gained the workspace/flag wiring (`mxfp8_gemm(..., split_k=, split_k_mode=)`). SEPARATE pending.
- Three `split_k` modes are supported:

| mode | fused reduction | run-to-run deterministic | performance | 
| ---- | --------------- | ------------------ | ----------- | 
| `SERIAL` | yes | yes | good |
| `PARALLEL` | yes | no | better |
| `SEPARATE` | no | yes | best |

### Performance against no splitK across the modes
image

Best speedup over the un-split kernel (split_k=1) across split-K {2,4,8,16,32}, per reduction mode (SERIAL / PARALLEL / SEPARATE), bf16 x bf16 -> fp32. X = M = N, Y = K; each cell is the speedup, annotated with the winning split factor (diverging scale centered at 1x: red = split-K slower, green = faster). GB200, CUDA graphs + L2 rotation, 500 iters.

### Performance against cuBLAS (splitK) baseline

image

### Performance against CUTLASS DSMEM Split-K Example

Best QuACK was selected over `split_k {1,2,4,8,16,32}` and `{SERIAL, PARALLEL, SEPARATE}`. CUTLASS was selected over `cluster_split_k {1,2,4,8,16}` from `dense_gemm_persistent_dynamic.py`.

`QuACK ms / CUTLASS ms`:

| K \ M=N | 64 | 128 | 256 | 512 | 1024 | 2048 |
|---:|---:|---:|---:|---:|---:|---:|
| 2K | 0.86 | 0.66 | 0.68 | 0.75 | 0.62 | 0.52 |
| 4K | 1.27 | 0.67 | 0.79 | 0.78 | 0.85 | 0.69 |
| 8K | 1.28 | 0.73 | 0.81 | 0.78 | 0.92 | 0.81 |
| 16K | 1.22 | 0.77 | 0.82 | 0.80 | 0.86 | 0.92 |
| 32K | 1.27 | 0.74 | 0.86 | 0.59 | 0.90 | 1.05 |
| 64K | 1.18 | 0.73 | 0.78 | 0.54 | 0.98 | 1.08 |
| 128K | 0.92 | 0.67 | 0.70 | 0.53 | 1.09 | 1.07 |
| 256K | 0.65 | 0.58 | 0.64 | 0.57 | 1.20 | 0.97 |

DSMEM reductions mainly won very small M/N `M=64` cases and a few large-M/high-K points; QuACK's best path was usually `SEPARATE`, often with `sk32` for long-K skinny/mid-sized shapes.

### Testing

`tests/test_gemm_split_k.py` runs each mode (and `split_k` 2/5, L 1/3, tf32/bf16/fp8 -> f32 out) end-to-end on GB200 and compile-only on SM90/SM120:
- Numerics vs an fp32 reference at the plain-kernel tolerance; `split_k=1` is bitwise identical to the baseline kernel; serial and staged f32 outputs agree bitwise.
- Run-to-run bitwise determinism (serial, staged); parallel is excluded by design.
- Full epilogue once on the sum: alpha, beta*C, rowvec + colvec bias, `add_to_output`, tensor-valued alpha/beta, and C aliasing the output (`gemm_add_inplace`, beta != 1).
- All three schedulers (non-persistent / static / CLC-dynamic) and m-major + mixed C/D majorness output paths.
- Edge cases: partial tiles, `split_k` > k-tiles (empty splits), cluster-rounded tile counts, odd epi-subtile counts (smem buffer rotation), staged grid not inflated, autotuner surface, and rejection of unsupported combos.
- Block-scaled MXFP8 split-K is covered in `tests/test_gemm_sm100_blockscaled.py` (SM100): SERIAL/PARALLEL across `split_k` 2/4 and batched, matching the plain MXFP8 kernel within ~1 bf16 ULP, serial deterministic, SEPARATE cleanly rejected.

### DIFF
--- a/quack/gemm_interface.py
+++ b/quack/gemm_interface.py
@@ -1,4 +1,5 @@
 # Copyright (c) 2025, Tri Dao
+from dataclasses import replace
 from typing import Optional, Tuple, Literal
 from functools import partial
 
@@ -6,7 +7,10 @@ import torch
 import torch.nn.functional as F
 from torch import Tensor
 
-from quack.gemm_config import GemmConfig, get_all_configs
+# SplitKMode is re-exported here as its canonical public import path; it is *defined*
+# in the gemm_config leaf because the kernel layer needs it at import time and this
+# module sits above quack.gemm in the import graph (importing it from here would cycle).
+from quack.gemm_config import GemmConfig, SplitKMode, get_all_configs
 
 from quack.autotuner import autotune, AutotuneConfig
 from quack.cute_dsl_utils import get_device_capacity
@@ -105,6 +109,15 @@ Activation = Literal[
     "geglu",
     "glu",
 ]
+
+
+def _check_split_k_unsupported(name: str, split_k: int) -> None:
+    """For use in interface wrappers that do not support splitK yet."""
+    if split_k not in (None, 1):
+        raise NotImplementedError(
+            f"{name} does not support split_k > 1; split_k is currently supported by "
+            "gemm, gemm_add, and gemm_add_inplace."
+        )
 
 
 def _concat_interleave(t):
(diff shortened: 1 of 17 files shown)


tridao (approved):

quack/utils.py
```diff
@@ -186,6 +186,61 @@ def warp_prefix_sum(val: Int32, lane: Optional[Int32] = None) -> Int32:
...
+    reuse, not the visibility of the gmem writes themselves. All lanes then sync
+    (ordering the issuing lane's completed work before the flag) and lane 0 performs
+    a gpu-scope release store.
+    """
+    if const_expr(drain_tma_store):
+        cute.arch.cp_async_bulk_wait_group(0)
+        cute.arch.fence_proxy("async.global")
+    cute.arch.sync_warp()
+    if cute.arch.lane_idx() == 0:
+        cute.arch.store(lock_ptr, value, sem="release", scope="gpu")
```
tridao, line 226: just confirming that this doesn't need threadfence?

quack/gemm_interface.py
```diff
@@ -194,12 +239,21 @@ def prune_invalid_gemm_configs(configs, named_args: dict, **kwargs):
...
         configs = [conf for conf in configs if not conf.kwargs["config"].use_tma_gather]
+    # Autotuned split-K: only the plain-gemm family exposes the knob; split_k=None means
+    # "tune the factor" (an explicit int is forced in gemm_tuned and needs no variants).
+    if (
+        "split_k" in kwargs
+        and kwargs["split_k"] is None
+        and not (varlen_m or varlen_k or gather_A)
+        and device_capacity in (9, 10, 11, 12)
+    ):
+        configs = _expand_split_k_configs(configs, kwargs["A"], kwargs["B"], device_capacity)
```
tridao, line 358: for sm90, is split-k supposed to work w pingpong as well? since autuner could pick that.
if not, we can avoid pingpong w split k and impl that later.

tridao:
Looks great, thanks so much @thakkarV . 
Just some minor questions to confirm

quack/gemm_interface.py
thakkarV: Yes, it does.

quack/utils.py
thakkarV: Copied the CUTLASS 2.x impl of serial reduction here, so I think we should not. @hwu36 @ionthruster can you confirm?

tridao:
feel free to merge when it's ready!

quack/utils.py
hwu36: correct. you don't need `threadfence` with `st.global.release.gpu`

cc @kerrmudgeon

PR #159, opened 2026-06-16
[GemmSm100] Add MXFP8 grouped GEMM (offs) dispatcher

santoshmo:
## Summary

Adds `mxfp8_grouped_gemm` — a per-expert **grouped** MXFP8 GEMM with a
`torch._scaled_grouped_mm`-style interface (2D-A / 3D-B + cumulative `offs`) — that
**routes onto the SM100 blockscaled kernels quack already ships** (dense batched-`L`
for uniform groups, varlen-M for ragged), plus one small, gated kernel addition
(`sfa_tile_aligned`) that lets the varlen-M path consume natural-layout scales.

Motivation: MoE needs a grouped MXFP8 matmul over a (often ragged) batch of experts.
`torch._scaled_grouped_mm` covers this on the PyTorch side; this gives quack a native
equivalent. On B200 it is **faster than `torch._scaled_grouped_mm` on both uniform
(~1.2–1.25×) and ragged 128-aligned (~1.06–1.09×) shapes** (numbers below), and is
correct across heavy skew and empty experts.

## API

```python
from quack.mxfp8_grouped_gemm import mxfp8_grouped_gemm, MXFP8GroupedGemm

# plain call, torch._scaled_grouped_mm-style
out = mxfp8_grouped_gemm(a, b, offs, sfa, sfb)

# sync-free fast path for equal groups (fixed-capacity MoE): route derived from
# shapes, offs never read on the host -> no device->host sync
out = mxfp8_grouped_gemm(a, b, offs, sfa, sfb, uniform=True)

# sync-free fast path for ragged 128-aligned groups: offs consumed only on device
out = mxfp8_grouped_gemm(a, b, offs, sfa, sfb, varlen=True)

# prepared instance (pre-packs the fixed B-scale + reuses the cached compile)
gemm = MXFP8GroupedGemm(b, sfb)        # b/sfb = fixed expert weights
out = gemm(a, offs, sfa)               # per step (offs may change per call)
out = gemm(a, offs, sfa, out=buf)      # reuse an output buffer
```

### Contract (mirrors `torch._scaled_grouped_mm`, 2D-A / 3D-B + `offs`)

| arg    | shape             | dtype             | notes                                                       |
| ------ | ----------------- | ----------------- | ---------------------------------------------------------- |
| `a`    | `(total_m, K)`    | `float8_e4m3fn`   | K-contiguous                                                |
| `b`    | `(E, K, N)`       | `float8_e4m3fn`   | K-contiguous (`b.transpose(-2, -1).is_contiguous()`)        |
| `offs` | `(E,)`            | `int32`           | **cumulative** per-group end rows; `offs[-1] == total_m`    |
| `sfa`  | `(total_m, K//32)`| `float8_e8m0fnu`  | raw per-block (1×32) scales, K-contiguous                   |
| `sfb`  | `(E, N, K//32)`   | `float8_e8m0fnu`  | raw per-block scales, K-contiguous                          |
| `out`  | `(total_m, N)`    | `bfloat16`        | optional, contiguous                                        |

Returns bf16 `(total_m, N)`.

## Routing

`offs` is inspected once (a single host sync), then dispatched to one of three paths
— all backed by existing kernels:

| case                                | route                                                         |
| ----------------------------------- | ------------------------------------------------------------ |
| uniform groups, all `g % 128 == 0`  | reshape to batched-`L` → dense `mxfp8_gemm_out`              |
| ragged, all boundaries `% 128 == 0` | varlen-M, **natural-layout SFA** (`sfa_tile_aligned`)         |
| ragged, non-128 boundaries          | pad each group up to a 128 multiple → batched-`L` dense       |

The varlen route packs SFA in **natural (unpadded)** layout
(`pack_scale_2d_to_blocked_contig`, exactly like torch's `to_blocked`) and the kernel
offsets per expert by `cu_seqlens[b] // 128` — no per-call dQaccum scatter (see below).

### Kernel change: `sfa_tile_aligned` (varlen-M)

`offset_batch_SFA` previously offset SFA by `cu_seqlens[b] // 128 + b` (dQaccum-padded);
the `+ b` slack exists so per-expert seqlens can be **arbitrary** (each rounds up to a
whole 128-tile). For **128-aligned** boundaries that slack is unnecessary, so a new
compile-time `sfa_tile_aligned` flag uses the natural offset `cu_seqlens[b] // 128` and
the caller passes SFA unpadded. Plumbed `GemmSm100 → offset_batch_SFA` via
`compile_blockscaled_gemm_tvm_ffi`; **default `False` preserves the existing dQaccum path,
and sm90/sm120 are untouched.** This removes the per-call scale-rearrange tax — the only
thing that kept ragged below torch — without touching the GEMM math (bit-identical; see
Correctness).

### Sync-free uniform fast path (`uniform=True`)

The per-call `offs`→host sync needed to route on the host is a full stream
synchronize: it gates the kernel launch on a host decision and injects a ~100 µs
launch bubble the GPU can't hide, which alone drops the plain call below torch. For
**equal groups** (fixed-capacity MoE) the route is *shape-determined* (`g = total_m
// E`), so `uniform=True` skips reading `offs` entirely — no sync, no bubble — and
runs the dense batched-`L` path directly. It is a bit-identical shortcut of the
default routed call.

### `MXFP8GroupedGemm` (prepared instance)

A callable that bakes the fixed B operand: the constructor pre-packs the B-scale and
the call reuses the cached compiled kernel, so each call does only the A-scale pack +
the kernel. Holding the packed B-scale as instance state (vs re-packing per call)
keeps it `>=` torch even at small group sizes, where the plain `uniform=True`
function's per-call B-pack would otherwise dominate a tiny kernel. Per-call work is
on the passed tensors, so it composes with CUDA-graph replay (call once on the
capture tensors, then update `a` / `sfa` / `out` in place between replays).

## Correctness

`tests/test_mxfp8_grouped_gemm.py`:

1. **Parity** — all 3 routes × `{eager, prepared}` × `{(256,256), (512,512), (6144,4096)}`,
   checked against a per-group **dequant** reference (`atol=rtol=1e-2`).
2. **Bitwise self-consistency** — `torch.equal(eager, prepared)` across all 3 routes,
   since both APIs compile the *same* `GemmSm100` kernel. This is strictly stronger
   than the tolerance check for catching a scale-pack / reshape plumbing regression
   in the prepared path, and it is robust (no dependence on split-K heuristics).
3. **Skew + empty experts** — heavy load imbalance and size-0 groups (empty experts,
   ubiquitous in MoE), across the varlen-M and padded routes; both APIs vs the dequant
   reference and bitwise `eager == prepared`.
4. **Uniform fast path** — `uniform=True` (function) and `MXFP8GroupedGemm(..., uniform=True)`
   are each bit-identical to the default offs-routed call.
5. **Varlen fast path** — `varlen=True` (function + class) matches the dequant reference and
   the offs-routed call; the function and class varlen paths are bit-identical.
6. **Kernel `sfa_tile_aligned` differential** (`tests/test_gemm_sm100_blockscaled.py`) — on
   128-aligned uniform / ragged / empty-expert / heavy-skew seqlens, the natural-SFA path is
   **bit-identical** to the dQaccum path and matches the dequant reference.

## Performance (B200 / SM100, `nvidia-cutlass-dsl` 4.5.2)

Kernel-only, **uniform** groups, both scales pre-packed (apples-to-apples), CUDA-event
timed, min-of-3 medians. The `torch._scaled_grouped_mm` baseline is validated correct
(within `1e-2` of a fp32 dequant reference) on every shape.

| E  | g (tok/expert) | K    | N    | quack kernel | torch._scaled_grouped_mm | speedup   |
| -- | -------------- | ---- | ---- | ------------ | ------------------------ | --------- |
| 8  | 512            | 4096 | 4096 | 2401 TF/s    | 1623 TF/s                | **1.48×** |
| 8  | 1024           | 4096 | 4096 | 2586 TF/s    | 2316 TF/s                | **1.12×** |
| 8  | 2048           | 4096 | 4096 | 2784 TF/s    | 2267 TF/s                | **1.23×** |
| 8  | 4096           | 4096 | 4096 | 2880 TF/s    | 2431 TF/s                | **1.18×** |
| 8  | 1024           | 6144 | 4096 | 2723 TF/s    | 2089 TF/s                | **1.30×** |
| 8  | 2048           | 6144 | 4096 | 2847 TF/s    | 2283 TF/s                | **1.25×** |
| 8  | 1024           | 7168 | 2048 | 2530 TF/s    | 2137 TF/s                | **1.18×** |
| 8  | 2048           | 7168 | 2048 | 2815 TF/s    | 2109 TF/s                | **1.33×** |
| 8  | 1024           | 2048 | 7168 | 2451 TF/s    | 2198 TF/s                | **1.12×** |
| 8  | 2048           | 2048 | 7168 | 2628 TF/s    | 2437 TF/s                | **1.08×** |
| 16 | 1024           | 4096 | 4096 | 2659 TF/s    | 2673 TF/s                | **0.99×** |
| 16 | 512            | 6144 | 4096 | 2551 TF/s    | 2343 TF/s                | **1.09×** |

**Geomean speedup: 1.19×** (range 0.99–1.48×; quack faster on 11/12, ~parity on the
largest E16 case).

**Varlen (ragged) route, end-to-end (K6144/N4096), vs torch ragged** (baseline validated
correct). The natural-SFA path makes ragged SFA prep just a `pack` (what torch pays for
`to_blocked`), removing the dQaccum rearrange tax → ragged now **beats torch**:

| ragged distribution | SFA prep | vs torch |
| ------------------- | -------- | -------- |
| balanced            | dQaccum (old)                    | 0.99× |
| balanced            | **natural (`sfa_tile_aligned`)** | **1.06×** |
| heavy-skew          | dQaccum (old)                    | 0.96× |
| heavy-skew          | **natural (`sfa_tile_aligned`)** | **1.09×** |

(Arbitrary non-128 seqlens still use the dQaccum path / the padded dense fallback.)

**End-to-end API perf (uniform).** The default plain call pays a per-call `offs`→host
sync — a full stream synchronize that gates the kernel launch and injects a ~100 µs
bubble — which caps it below torch. The sync-free paths recover the kernel win
(all relative to `torch._scaled_grouped_mm` = 1.0):

| E×g (K/N=4096, *=K6144) | plain (routed) | `uniform=True` | `MXFP8GroupedGemm` (uniform) |
| --- | --- | --- | --- |
| 8×512   | 0.50× | 0.77× | **1.16×** |
| 8×1024  | 0.56× | 1.01× | **1.23×** |
| 8×2048* | 0.81× | 1.08× | **1.25×** |
| 16×1024 | 0.71× | 1.02× | **1.25×** |

`uniform=True` removes the sync (≥ torch for g ≥ 1024); the prepared instance also
pre-packs the B-scale, so it stays ≥ torch including small g. These end-to-end numbers
are noisier than the kernel-only table (~20% run-to-run, no clock pinning) and vary with
hardware / driver / cutlass-dsl version — treat as indicative.

## Bitwise note (correctness, not a guarantee)

On the uniform shapes we compared directly with `torch.equal` (K = 512–6144, several
per-group sizes), the output is **bit-identical** to `torch._scaled_grouped_mm` — even
though quack is faster. Both lower to the same tcgen05 blockscaled MMA; with enough M
to fill the SMs **neither uses split-K**, so K is accumulated in the same sequential
fp32 order with a single bf16 epilogue round → identical bits. The speedup is
pipelining/scheduling, which does not change the accumulation *order*.

This is **not asserted** in the tests on purpose: it is fragile by nature — split-K
(e.g. small-M / large-K shapes) tree-reduces in a different order, and any tile /
cluster / kernel-version change breaks it. So the cross-kernel test stays at `1e-2`
tolerance, and only the `eager == prepared` invariant (same kernel) is pinned bitwise.

## Implementation notes

- New files: `quack/mxfp8_grouped_gemm.py` + `tests/test_mxfp8_grouped_gemm.py`.
- Small **gated** kernel change: `quack/varlen_utils.py` (`offset_batch_SFA` +
  `sfa_tile_aligned`), `quack/gemm_sm100.py` (config attr + call site),
  `quack/blockscaled_gemm_utils.py` (`compile_blockscaled_gemm_tvm_ffi` flag), and a
  differential test in `tests/test_gemm_sm100_blockscaled.py`. Default off; sm90/sm120
  untouched.
- Reuses `mxfp8_gemm_out`, `_compile_cached`, `_default_tiler_cluster`,
  `compile_blockscaled_gemm_tvm_ffi`, `pack_scale_2d_to_blocked_contig`,
  `scale_view_for_kernel`.
- The dense route passes **raw** e8m0 scales (the interface packs internally); the
  varlen route passes **natural pre-packed** scales directly.
- The varlen kernel is `lru_cache`'d on `(n, k, e, tiler, cluster, sfa_tile_aligned)` —
  `cute.sym_int` abstracts `total_m`, so one compile serves all group configs of that shape.

## Limitations / future work

- Arbitrary **non-128** per-expert seqlens still use the dQaccum path / padded dense
  fallback; the natural-SFA fast path requires 128-aligned boundaries (caller-asserted
  via `varlen=True`, or detected by the default router).
- The default plain call still pays one `offs`→host sync to *route*; use `uniform=True` /
  `varlen=True` (or the `MXFP8GroupedGemm` instance) to avoid it when the regime is known.
- `mxfp8_gemm_out` exposes only `mma_tiler_mn` / `cluster_shape_mn`; finer epilogue /
  staging knobs could add a few % on some shapes.
- Requires `nvidia-cutlass-dsl >= 4.5.2` (uses `OperandMajorMode`; the 4.5.x scale
  layout, `(l, rm, rk, 512)`).

### DIFF
--- a/quack/blockscaled_gemm_utils.py
+++ b/quack/blockscaled_gemm_utils.py
@@ -608,6 +608,7 @@ def compile_blockscaled_gemm_tvm_ffi(
     use_clc_persistence: bool = True,
     varlen_m: bool = False,
     varlen_k: bool = False,
+    sfa_tile_aligned: bool = False,
 ) -> Callable:
     """Compile the SM100 blockscaled GEMM.
 
@@ -625,6 +626,7 @@ def compile_blockscaled_gemm_tvm_ffi(
         GemmDefaultSm100,
         sf_vec_size=sf_vec_size,
         use_clc_persistence=use_clc_persistence,
+        sfa_tile_aligned=sfa_tile_aligned,
     )(cutlass.Float32, ab_dtype, mma_tiler_mn, (*cluster_shape_mn, 1))
     compile_epi_args = gemm.EpilogueArguments()
     scheduler_args = make_scheduler_args(
--- a/quack/gemm_sm100.py
+++ b/quack/gemm_sm100.py
@@ -142,6 +142,7 @@ class GemmSm100(GemmTmaBase):
         use_clc_persistence: bool = True,
         concat_layout: tuple | None = None,
         use_pdl: bool = True,
+        sfa_tile_aligned: bool = False,
     ):
         """Initializes the configuration for a Blackwell dense GEMM kernel.
 
@@ -168,6 +169,10 @@ class GemmSm100(GemmTmaBase):
         self.acc_dtype: Type[cutlass.Numeric] = acc_dtype
         self.sf_vec_size = sf_vec_size
         self.blockscaled = sf_vec_size is not None
+        # When True, SFA is in natural (unpadded) packed layout and per-expert offsets
+        # are computed as cu_seqlens[b] // 128 (valid only for 128-aligned varlen_m
+        # boundaries); when False, SFA is dQaccum-padded (cu // 128 + b).
+        self.sfa_tile_aligned = sfa_tile_aligned
         assert len(mma_tiler_mnk) in [2, 3], "MMA tiler must be (M, N) or (M, N, K)"
         valid_2cta_m = (128, 256) if not self.blockscaled else (256,)
         self.use_2cta_instrs = cluster_shape_mnk[0] % 2 == 0 and mma_tiler_mnk[0] in valid_2cta_m
@@ -1115,7 +1120,9 @@ class GemmSm100(GemmTmaBase):
                     # the A-data offset — allows varlen_m seqlens that aren't
                     # multiples of 128.
                     gSFA_mkl = cute.local_tile(
-                        varlen_manager.offset_batch_SFA(mSFA_mkl, batch_idx),
+                        varlen_manager.offset_batch_SFA(
+                            mSFA_mkl, batch_idx, tile_aligned=self.sfa_tile_aligned
+                        ),
                         cute.select(self.mma_tiler, [0, 2]),
                         (mma_tile_coord_mnl[0], None),
                     )
--- a/quack/mxfp8_grouped_gemm.py
+++ b/quack/mxfp8_grouped_gemm.py
@@ -1,0 +1,247 @@
+# Copyright (c) 2026, Tri Dao.
+"""MXFP8 grouped GEMM dispatcher for quack.
+
+Routes a grouped (per-expert) MXFP8 GEMM over cumulative ``offs`` onto quack's
+two existing SM100 blockscaled engines, mirroring ``torch._scaled_grouped_mm``:
+
+  * uniform 128-aligned groups -> batched-L dense (:func:`mxfp8_gemm_out`)
+  * ragged 128-aligned groups  -> varlen-M (:func:`compile_blockscaled_gemm_tvm_ffi`)
+  * ragged non-128 groups      -> pad to a 128 multiple, batched-L dense
+
+Contract (2D-A / 3D-B + offs, like ``torch._scaled_grouped_mm``)::
+
+    a:    (total_m, K)     float8_e4m3fn, K-contiguous
+    b:    (E, K, N)        float8_e4m3fn, K-contiguous (b.transpose(-2,-1).is_contiguous())
+    offs: (E,)             int32, CUMULATIVE per-group end rows; offs[-1] == total_m
+    sfa:  (total_m, K//32) float8_e8m0fnu, K-contiguous (raw per-block scales)
+    sfb:  (E, N, K//32)    float8_e8m0fnu, K-contiguous (raw per-block scales)
+    out:  optional (total_m, N) bfloat16, contiguous
+
+Returns bf16 ``(total_m, N)``.
+"""
+
+from functools import lru_cache
+from typing import List, Optional, Tuple
+
+import cutlass
+import torch
+from torch import Tensor
+
+from quack.blockscaled_gemm_utils import (
+    compile_blockscaled_gemm_tvm_ffi,
+    pack_scale_2d_to_blocked_contig,
+    scale_view_for_kernel,
+)
+from quack.gemm_blockscaled_interface import (
+    _compile_cached,
+    _default_tiler_cluster,
+    mxfp8_gemm_out,
+)
+
+SF_VEC = 32
+
+
+def _uniform_group_size(offsets_host: Tuple[int, ...]) -> Optional[int]:
+    prev, g = 0, None
+    for cur in offsets_host:
+        sz = cur - prev
+        if g is None:
+            g = sz
+        elif sz != g:
+            return None
+        prev = cur
+    return g
+
+
+def _group_sizes(offsets_host: Tuple[int, ...]) -> List[int]:
+    prev, out = 0, []
+    for cur in offsets_host:
+        out.append(cur - prev)
+        prev = cur
+    return out
+
+
+def _boundaries_128_aligned(offsets_host: Tuple[int, ...]) -> bool:
+    prev = 0
+    for cur in offsets_host:
+        if prev % 128 != 0 or cur % 128 != 0:
+            return False
+        prev = cur
+    return True
+
+
+def _dqaccum_padded_sfa(sfa: Tensor, group_sizes: List[int], sf_k: int, e: int) -> Tensor:
+    """Pack SFA for the varlen-M kernel in dQaccum-padded layout: expert i's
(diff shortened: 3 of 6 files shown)


santoshmo:
@tridao @thakkarV should I include benchmarks against the existing varlen manager path?

PR #160, opened 2026-06-24
Allow boolean boolean tensors for make_fake_tensor

drisspg:
# Summary

Before if you tried to convert a boolean tensor to fake it would use numbits = 1 so 1//8 = 0 and min alignment would fail in lowering


quack/compile_utils.py
```diff
@@ -14,6 +14,5 @@ def make_fake_tensor(dtype, shape, divisibility=1, leading_dim=-1) -> Optional[c
         cute.sym_int64(divisibility=divisibility) if i != leading_dim else 1
         for i in range(len(shape))
     )
-    return cute.runtime.make_fake_tensor(
-        dtype, shape, stride=stride, assumed_align=divisibility * dtype.width // 8
-    )
+    assumed_align = divisibility * max(dtype.width // 8, 1)
```
tridao: when e.g. divisiblity = 4, we'd have alignment = 4 bytes? I feel it should be 1 byte instead?
maybe we do ceil_div(divisibility * dtype.width, 8)? 
We should think about the case of fp4 (dtype.width = 4) as well.

tridao:
Thanks for the fix — the bug is real (`Boolean.width` is 1 *bit*, so `divisibility * width // 8` rounds to 0). I pushed to your branch: merged main (the file had drifted) and adjusted the formula slightly. `max(dtype.width // 8, 1)` treats every sub-byte dtype as byte-per-element, which is right for Boolean (DLPack kDLBool is 1 byte/element) but would over-claim alignment 2x for genuinely bit-packed int4/fp4 (div=32 should be 16B, not 32B). Now it special-cases Boolean's storage width:

```python
storage_bits = 8 if dtype is cutlass.Boolean else dtype.width
assumed_align = max(divisibility * storage_bits // 8, 1)
```

Also extended the test to run a real bool tensor through TVM-FFI and check values, plus pin the int4/fp4 alignment. Will merge once CI is green.

quack/compile_utils.py
tridao: Resolved as 1 byte in 8ed0a5b — no Boolean special case, just `max(divisibility * dtype.width // 8, 1)`. Rationale: generic call sites (e.g. epilogue ops' divisibility=4) pick divisibility for wide dtypes' vectorization, so a bool tensor flowing through would inherit a base-alignment promise nobody validated (a mask sliced at an odd element offset sits at an odd byte). Went with floor rather than ceil_div since ceil over-claims non-divisible sub-byte cases (fp4 div=3 → 2B for 1.5 bytes of data); fp4/int4 divide down correctly either way (div=32 → 16B, pinned in the test).

tridao (approved):

PR #169, opened 2026-07-09
[GEMM] BlockScaledOperand API for block-scaled operands + mixed fp8 A/B

thakkarV:
Introduces typed `BlockScaledFormat` and `BlockScaledOperand` APIs, replacing legacy `tuple(data, scale)` blockscaled GEMM operands.

- Adds quantization, dequantization, transpose, device movement, serialization, and torch.compile/pytree support.
- Plumbs independent A/B formats through GEMM, add, activation, autotuning, epilogues, custom ops, and TVM FFI.
- Automatically folds NVFP4 per-tensor scales into alpha and validates operand orientation, layouts, devices, packing, and MMA compatibility.
- Enables mixed MXFP8 E4M3/E5M2 SM100 GEMMs and adds byte-container MXFP4/MXFP6 groundwork with explicit kernel capability gates.
- Reserves the blockscaled-output API and documents future scale-recipe, layout, and SF- generation extensions.
- Adds comprehensive numerical coverage for formats, orientations, varlen/batched paths, serialization, compiler fullgraph paths, and validation failures.


tridao:
Pushed a cleanup commit (4a29e03) directly to the branch — no design changes, just tightening:

- **Hard break confirmed**: we're keeping the TypeError on tuples (no deprecation cycle). Updated the PR description, which still said "DeprecationWarning (removal >= v0.8)".
- Removed `summary.md` — that's the draft description of the stacked mixed-input follow-up PR, leaked into this one.
- `_packed_dim`: size-1 dims report arbitrary strides, so a `(Kp, 1)` K-major fp4 operand can present stride 1 on both dims; the scan picked the size-1 dim and falsely rejected `quant_dim=-2`. Now prefers extent>1 dims; regression test added.
- The tcgen05 kind rules live in three places (`mma_kind_for_pair`, `_blockscaled_mma_inst_k`, `is_valid_dtypes_and_scale_factor_vec_size`) — added cross-referencing comments and `test_mma_kind_mirrors_kernel_inst_k` (host-only) pinning the torch-layer/kernel-layer mirror so they can't drift.
- Fixed stale references: the fp6 "no quantizer" parenthetical in `quantize()` (fp6 quantizers exist — only e5m2 lacks an encoder), "milestone M5 / section 8" → section 7, "byte-container fp4 is backlog" → `mxfp4_byte` exists, and the design doc's cache-key paragraph (both compile paths already key on `a/b_mma_dtype`, so the fp6 e2m3/e3m2 collision it warns about can't happen).

Verified: the kind-mirror test passes host-only; the `_packed_dim` fix verified on CPU (construction + dequant match, and the old scan demonstrably picks the wrong dim). The CUDA-side container tests are queued to re-run once our runner node recovers (fabric manager is down).

tridao:
Node recovered — full blockscaled suite run on the branch with the cleanup commit: `test_blockscaled_operand.py` 31 passed, `test_gemm_blockscaled_interface.py` 86 passed, `test_gemm_sm100_blockscaled.py` 91 passed / 14 skipped (B300). The two async-compile pool errors are the negative tests' keys (gated fp6×fp6 and fp8×byte-fp4 combos correctly rejected by the per-arch gate).

tridao:
Pushed a second commit: the follow-up design for generalizing this API beyond MX/NVFP4, plus the three seam hardenings it justifies **in this PR** (design doc: `AI/blockscaled_recipes.md`; new "Extension seams" section in `AI/blockscaled_api.md`).

**The design in one paragraph.** A block-scaled operand is `decode(qdata) · S[i//bm, j//bk] · (pts)` — a two-integer block grid, a scale dtype, and at most two levels. `sf_vec_size` is the K-extent of a `(1, bk)` grid; an `sf_block_mn` extent generalizes it to DeepSeek-style 2D (128×128) weight scales. Three axes get three homes: **element encoding + scale recipe** on the format (the interchange contract), **physical scale layout** on the operand (like strides — the blocked `(rm, rk, 32, 4, 4)` atom is the tcgen05 consumption layout, and the *linear* `(MN, K/vec)` layout becomes a per-operand tag, never inferred from rank), and **pair legality** in the kind function, where the tcgen05 kinds become a subset alongside a software `blockwise_promote` kind (DeepGEMM-style fp8 1d1d/1d2d, and one-sided bf16 + per-K-block fp32 scales — sided-ness is per-kind, not an interface invariant). Driving consumers: an SM90 blockwise-promotion GEMM and the SM100 linear-SF (cp.async) loader, both in flight on our side; the generalization itself lands with whichever kernel merges first, since new fields with no consumer are exactly what this PR trimmed.

**What this commit changes in code** (deliberately minimal):

- `cutlass_dtype_name` is now `Optional[str]`: `None` marks a format with no CuTe-DSL element type — host-side complete (quantize/dequantize/serialize), consumable by a kernel only once it declares its own (copy dtype, MMA dtype, convert) triple. `from_cutlass_dtypes` skips `None` entries, so registering such a format can no longer crash registry-wide dtype lookup. (We verified the DSL side empirically: the bundled MLIR even ships `f8E3M4`, but the DSL's reverse type dispatch and every dtype-keyed table don't know unregistered types — and torch can't carry them across dlpack anyway — so typeless-format + kernel-declared triple is the honest model, not a workaround.)
- `mma_kind_for_pair` no longer falls through to `mxf8f6f4` blindly: it gates on the MMA element class (`Float8*/Float6*/Float4*` from the DSL type name), so a format whose elements no tcgen05 kind can consume fails at kind selection with a clear error instead of at the per-arch gate (or worse, a `getattr` crash) later.
- Test: `test_format_without_dsl_element_type` pins all three behaviors.

Everything else in the recipes doc is design-only for now. Tests are green on B300 after the change (operand 32, interface 86, kind-mirror pass).

tridao:
Pushed `e7bf9c2`, restructuring the design material from the previous commit:

- **One design doc**: `AI/blockscaled_recipes.md` is folded into `AI/blockscaled_api.md` as sections 9–10 and deleted. §9 is the generalization design (scale recipes as a `(bm, bk, scale_dtype)` grid — the factored form of cuBLASLt's `cublasLtMatmulMatrixScale_t` / torch's `_ScalingType` enums; per-operand scale layouts; software kinds with per-kind sided-ness; the prepared-operands fence for weight-only repack blobs). §10 is a stress-test table that doubles as the abstraction's regression suite: seven adversarial cases (e3m4, int4/nvint4, DeepSeek 1×128/128×128, bf16-with-K-block-scales, linear SF layout, W4A16 nvfp4 reuse, W4A8/W8A16 rowwise) producing just two amendments — a full-extent `sf_vec_size` sentinel and "k-invariant scales compose freely".
- **Worked examples in code**: commented future-format rows next to `BLOCKSCALED_FORMAT_REGISTRY` and their executable form, `test_future_recipe_examples`, which pins that each case fails at the *kind* layer with the right reason until its consumer kernel lands.
- **One behavior fix the examples flushed out**: `mma_kind_for_pair` gated the MMA element class but not the scale *recipe*, so a DeepSeek-recipe fp8 format (valid elements, fp32/vec-128 scales) would have silently claimed `kind::mxf8f6f4`. It now checks both axes (e8m0/vec-32 required for the mxf8f6f4 tail), and the element/recipe gate runs before the packed-storage check so packed non-fp4 formats get "no MMA element type" instead of fp4-specific requantize advice.

No API changes; the generalization fields themselves (`sf_block_mn`, `scale_layout`, software kinds) still land with their first consumer kernel per §9.8. Full suite re-run green on B300: operand 34, interface 86, kind-mirror pass.

tridao (approved):

tridao:
Rebased onto current `main` (13 commits, including the epilogue redesign) and force-pushed; the four commits are otherwise unchanged in content.

The substantive part of the rebase: `main`'s epilogue redesign moved `gemm_act`/`gemm_gated` onto the epilogue-object surface (`linear_act_mod` → `EpiMod` → `gemm_host` plans), which replaced the old `gemm_act_tuned`/`_gemm_act_execute` path this PR had threaded formats through. The blockscaled format plumbing is re-expressed on the new path:

- `bs_format_a`/`bs_format_b` thread through `EpiMod.__call__`/`gemm`/`plan`, `_iface_execute`, the epi autotuner (`tuned_mod_gemm` + tuner key), and `build_gemm_epi_plan`/`_compile_gemm_epi` → `compile_gemm_kernel` (the `a/b_mma_dtype` split now reaches the epilogue compile path too, keyed by `jit_cache`).
- The `quack::gemm_epi` compile-path op carries the format names in its meta string, with e8m0 scales crossing the mutable-op boundary as uint8 views and decoded format-driven in the op body — closing the same Inductor `decompose_auto_functionalized` gap the `gemm_out` seam already handled.
- `EpiMod.gemm` resolves the descriptors once (`resolve_blockscaled_formats`) and passes `fmt_a`/`fmt_b` to `validate_blockscaled_sf`, same as the `quack.gemm` path.

Re-verified on B300 after the rebase: blockscaled suite green (operand 33, interface 86, sm100 91 passed / 14 skipped), and the non-blockscaled epilogue surface unregressed (`test_epilogue_iface` 19 passed / 1 xfailed, `test_gemm_epilogue -k "act or gated"` 11 passed).

PR #170, opened 2026-07-10
[GEMM] Full blockscaled MXFP4/6/8 QMMA and MXFP4/NVFP4 OMMA support for SM100 GEMM

thakkarV:
Stacked on #169

- Add native SM100 mixed MXFP8/MXFP4/MXFP6 GEMMs using U4/U6 TMA unpack into byte-domain shared memory.
- Introduce packed MXFP6 formats, quantizers, logical/storage shape mapping, and bit-exact legacy conversion.
- Preserve existing FP6 APIs, pickles, and pytrees with versioned, JSON-safe serialization and migration.
- Strengthen tile, instruction/atom K shape, format, pointer, and stride validation across dispatch and cached launches.
- Extend mixed-format benchmarks, varlen-M support, documentation, and numerical regression coverage.

Every A/B pair drawn from `{mxfp8_e4m3, mxfp8_e5m2, mxfp4, mxfp6_e2m3_packed, mxfp6_e3m2_packed}` is now supported natively on SM100. Mixed pairs use tcgen05 `kind::mxf8f6f4`; homogeneous MXFP4 uses `kind::mxf4nvf4` path. NVFP4 remains restricted to NVFP4 x NVFP4.

Sub-byte operands in `mxf8f6f4` pairs stay packed in global memory and are expanded through the U4/U6 TMA unpack tensor maps (`CU_TENSOR_MAP_DATA_TYPE_16U4/16U6_ALIGN16B`). This includes FP6 x FP4, where both operands use the unpack path.

### TMA Unpack

- The TMA builder receives `internal_type=cutlass.Uint8` while the global-memory
  tensor retains, or is reinterpreted to, its true sub-byte element type.
  Recasting the source itself to `Uint8` would select a plain U8 tensor map and
  disable unpacking.
- Shared-memory layouts, `MemRange` storage, and stage budgeting use byte-domain
  `Uint8`, matching CUTLASS's `SmemAllocType = uint8_t` convention. Each group
  of 16 packed elements expands to a 16-byte shared-memory footprint.
- The `Uint8` shared-memory tensor feeds `make_fragment_A/B` directly. Its width
  and strides produce the correct MMA descriptor offsets; an FP4-domain layout
  would instead encode the packed `mxf4` convention.
- The mbarrier transaction count still uses packed source bytes, matching
  CUTLASS's `sizeof_bits<Element> * cosize(SmemLayout)` calculation.

Unpack operands must be K-major, have logical `K % 128 == 0`, and satisfy
32-byte base and non-unit stride alignment. These checks run during planning and
again at launch so cached plans cannot bypass pointer validation. Explicit tile-K
values must also contain complete scale-factor chunks.

### Packed FP6 and Compatibility

PyTorch has no packed FP6 dtype, so FP6 crosses TVM FFI as a raw `uint8` tensor
with shape `(..., 3K/4)`. The bytes contain a little-endian 6-bit stream matching
CUTLASS `SubbyteReference`; the kernel reinterprets this storage as the selected
FP6 element type. `a_mma_dtype` and `b_mma_dtype` carry that distinction through
both compilation paths.

The existing `mxfp6_e2m3` and `mxfp6_e3m2` names retain their byte-container
semantics for API and checkpoint compatibility. Kernel-ready storage uses the
explicit `*_packed` formats, and `BlockScaledOperand.to_packed()` converts legacy
operands without requantization. `mxfp4_byte` is likewise retained as a host-side
compatibility format but rejected by GEMM.

`BlockScaledFormat` keeps the positional `elems_per_container` field for
compatibility and adds versioned storage-layout metadata plus `storage_k()` and
`logical_k()` helpers for FP6's four-elements-per-three-bytes mapping. Legacy
pickles and pytrees remain loadable, while new pytree contexts are versioned and
JSON-safe.

### Unified FP4 Atom

Both MXFP4 and NVFP4 homogeneous pairs now construct `MmaMXF4NVF4Op`. This
models `kind::mxf4nvf4` as one atom parameterized by its scale configuration:

- MXFP4: E8M0 scales, vector size 32
- NVFP4: E4M3 scales, vector size 16

The wrapper removes the DSL class's hardcoded vector size. The backend continues
to derive the final PTX spelling from the scale configuration, so
`mma_kind_for_pair` no longer models MXFP4 as a separate atom.

## Testing

- Added a 10-pair mixed-format numerical matrix covering FP8, packed FP4, and
  both packed FP6 encodings.
- Added rejection tests for invalid K granularity, tile K, base alignment, and
  row stride, plus numerical tests for valid padded layouts.
- Added bit-exact FP6 packing, eager/compiled quantizer parity, dequantization,
  serialization migration, and byte-to-packed conversion tests.
- Extended varlen-M coverage to E5M2 and packed FP6 and revalidated MXFP4/NVFP4
  through the unified FP4 atom.
- Full suite: `pytest tests/ -n 8 --async-compile=32`. Remaining failures were
  limited to the pre-existing non-blockscaled BF16 `test_linear_varlen_m`
  multiprocessing flake, which reproduces on `main` and passes in isolation,
  plus environment-dependent async-compilation infrastructure tests.

### DIFF
--- a/AI/blockscaled_api.md
+++ b/AI/blockscaled_api.md
@@ -77,7 +77,6 @@ constraints anchor the rationale where noted.
   activations or accept dim1-requant error (the standard MX tradeoff).
 - **NVFP4 outputs with a computed global amax** (needs a delayed or two-pass amax;
   section 7 requires an explicitly provided per-tensor scale instead).
-- **An e5m2 `to_mx` encoder** (`from_parts` is the e5m2 construction path).
 
 ## 3. `BlockScaledFormat` and `BlockScaledOperand` (`quack/blockscaled/operand.py`)
 
@@ -110,8 +109,8 @@ class BlockScaledFormat:
   GEMM can execute a given (A, B, D) dtype combination is asserted per-architecture
   inside the `gemm_smXXX` classes (SM100 accepts any fp8/fp6/fp4 mix under
   kind::mxf8f6f4 - see section 6).
-- E5M2: descriptor + `from_parts` work (the kernel admits e5m2); `quantize` raises
-  because `to_mx` has no e5m2 encoder - `from_parts` is the e5m2 construction path.
+- E5M2: `quantize` encodes via `to_mx(..., elem_dtype=torch.float8_e5m2)`
+  (fp8_max 57344, max_pow2 15); `from_parts` also works for pre-quantized data.
 - `mxfp4_byte` and unversioned byte-container MXFP6 checkpoints retain their logical
   shape and dequantization after load. They are host-side compatibility formats only;
   `BlockScaledOperand.to_packed()` preserves their codes/scales exactly for GEMM use.
@@ -443,7 +442,7 @@ but that is the D5 trap again and breaks on the first same-rank layout.
   (e8m0 keeps the floor rule). bf16 and int-group formats RAISE - floating-
   point relative precision is scale-invariant and group scales are upstream-
   structural (GPTQ/AWQ groups, folded norms), so `from_parts` is the
-  construction path (the e5m2 precedent).
+  construction path.
 
 ### 9.4 Scale layouts (per-operand)
 
--- a/AI/blockscaled_api.md
+++ b/AI/blockscaled_api.md
@@ -30,18 +30,22 @@ constraints anchor the rationale where noted.
 | D1 | **The container is a non-differentiable plain object** - a frozen dataclass, not a `torch.Tensor` subclass, so it structurally cannot enter autograd. Training on quantized operands is out of scope by design: a forward-quantized operand is scaled along the wrong axis for both backward GEMMs (dgrad contracts N, wgrad contracts M; scales block along K), and the square-weight case (K==N) fails *silently*. Any training story is hp master weights + per-GEMM requantization, layered above this container. |
 | D2 | **`shape`/`dtype` report the logical hp view** (unpacked shape, `orig_dtype`, default bf16) as honest computed properties. Storage truth lives in `qdata`/`scale`. The container is unpacked at the top of each public entry point (where `_unpack_operand` sits); **nothing below that line ever sees it** - custom ops, autotuner, `jit_cache`, TVM-FFI all see flat plain tensors. |
 | D3 | **Transpose = qdata stride-swap view; scale untouched.** The blocked `(rm, rk, 32, 4, 4)` scale atom is not view-transposable; scale is defined as "always blocked along K in either orientation". Preserves the load-bearing `weight.T` / `qw.mT` idiom. qdata<->scale shape coupling is validated at gemm dispatch time (`validate_blockscaled_sf`), NOT at construction - varlen padded scale buffers must stay constructible. |
-| D4 | **Format = frozen dataclass descriptor** (`BlockScaledFormat`), not an enum: storage dtype, elem bits, elems-per-container, scale dtype, sf_vec_size, k-alignment are *data per format*. New formats are data, not code. Registry + singletons; legacy string names (`"mxfp8"`) coerce. |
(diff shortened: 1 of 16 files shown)


tridao:
Very nice — this matches the design we'd prototyped on a WIP branch almost exactly (same conclusions independently: `internal_type` selects the U4/U6 unpack tensormaps, all smem accounting in the byte domain, inst K = 32 for the unpacked kinds, txn bytes = packed gmem bytes), and the packed-fp6 FFI story + per-launch pointer validation here go further than what we had. I pushed four commits onto this branch pulling in the deltas our branch had that this PR lacked, plus one finding from review:

- **e3c8441** — a real MXFP8-E5M2 encoder: `to_mx(..., elem_dtype=torch.float8_e5m2)` (fp8_max 57344 / max_pow2 15, rceil + floor), registered as the `mxfp8_e5m2` quantizer. This removes the three quantize-as-e4m3-then-value-cast-the-codes workarounds (benchmark, varlen-M generator, interface tests) and enables e5m2 through `BlockScaledOperand.quantize` and the `init=quant` generators.
- **8ed14c0** — varlen_k coverage for mixed fp8 A/B (`b_dtype` on the generator + a public-API test, e4m3 x e5m2 both orders, aligned and ragged K). Mixed fp8 varlen_k already worked; it was just untested.
- **2101d68** — fp4 operand-generator fix: the `(mn, k/2, l)` contiguous allocation puts stride 1 on L instead of K for l > 1 (latent on main since every fp4 correctness case is l=1). Now allocates `(l, mn, k/2)` + permute, pinned by a batched l=2 mxfp4 case that fails with the old allocation.
- **d1d12b2** — one review finding, docs-only: `_reinterpret_packed_fp6`'s docstring claimed the non-K stride conversion (`* 4 // 3`) is exact, but it isn't for pitches that aren't multiples of 3 — your own padded test uses a 416 B pitch → 554.67 fp6 elements, floored to 554 = 415.5 B. It's still correct: I probed pitches 416/448/544/4128 with poisoned padding (and with the floor patched to a ceil) and everything is bit-exact, so the tensormap encode rounds element strides to the nearest TMA granule and the < 0.75 B residue is always recovered — *because* validation restricts strides to 32 B multiples. That makes the 32 B check load-bearing for addressing correctness, not just for the tensormap's alignment rule, so I documented the invariant and committed the probe (`AI/probe_fp6_pitch.py`).

Verification on B300: the three blockscaled test files pass in full (271 passed / 14 skipped), including all 10 mixed pairs through the new encoder; benchmark ref-check passes for `--bs_format_a mxfp8_e5m2 --bs_format_b mxfp4`.

Two questions / possible follow-ups, neither blocking:

1. **Compat machinery size**: the `__setstate__` migration for the "interim packed-FP6 schema", the `*_byte` aliases, and the interim-name canonicalization in `_coerce_format` exist to migrate pickles from a schema that never shipped on this repo. If those checkpoints only exist on your side, most of that surface could be dropped and the PR gets meaningfully smaller — happy either way, just checking it's intentional.
2. **MN-major sub-byte under mxf8f6f4**: this PR forbids non-K-major sub-byte operands outright. The hardware/tensormap path admits them (with the contiguous dim % 128 rule, including the 2-CTA per-CTA-tile halving); our branch had the `can_implement` plumbing for m-major fp4 A / n-major fp4 B but we never validated it numerically. Fine to leave out here — flagging it as a known extension if a wgrad-shaped use case shows up.

thakkarV:
@tridao the four commits you mention seem to be orphaned?

tridao:
Re-pushed them — a force-push reset the branch to 84aeb92 (its state before my commits), which orphaned e3c8441..d1d12b2. The branch head was identical to what I'd based on, so restoring was a clean fast-forward; the tip is d1d12b2 again. If you have local work in flight, `git pull` (or rebase onto d1d12b2) before your next push so we don't clobber each other.

tridao (approved):

PR #172, opened 2026-07-13
[Epi] Expose pre-activation alpha through `gemm_act`

thakkarV:
The default-epi GEMM already plumbs alpha end-to-end, but the act/gated path never exposed it: linear_act_mod builds its own epilogue fn without the Scalar("alpha") op, so callers needing `act(alpha * A @ B)` had to materialize a pre-scaled copy of A.

- `epilogues.linear_act_mod` has_alpha: pins Scalar("alpha") and prefixes the generated body with `x = acc * alpha`, so the scale applies to the accumulator BEFORE C / rowvec / colvec / activation (same math order as the default epilogue).
- `gemm_act` / `run_gemm_act_plan` take `alpha (float | Tensor);` the neutral 1.0 folds to the alpha-free epilogue.
- `gemm_interface`: `gemm_act_tuned` / `gemm_gated_tuned` / `_gemm_act_execute` pass it through; the public `gemm_act` keys the eager plan on `scalar_mode(alpha)`. gemm_act_out / gemm_gated_out grows the `alpha`/`alpha_tensor` schema split used by `gemm_out`; `gemm_act_ref` forwards alpha to `gemm_ref`/`gemm_add_ref`.

Tested on GB200 (SM100): new `test_gemm_act_alpha` covers float and tensor alpha x {tanh, relu} x bias on/off against the alpha-carrying reference.

### DIFF
--- a/quack/epilogues.py
+++ b/quack/epilogues.py
@@ -412,12 +412,16 @@ _SR_OPS = (Scalar("sr_seed", dtype=Int32),)
 
 
 @functools.lru_cache(maxsize=None)
-def linear_act_mod(activation, *, gated, has_c, has_rowvec, has_colvec, sr=False):
-    """gemm_act/gemm_gated as a mod: D = acc (+ C + rowvec + colvec), aux =
-    act(D). Math order matches apply_linear_epilogue: C, rowvec, colvec."""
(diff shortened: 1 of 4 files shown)


tridao (approved):

tridao:
Looks good, thanks @thakkarV !

PR #173, opened 2026-07-16
[Gemm] Add Gemm + ReduceScatter/AllReduce

Nicyzk:
# GEMM + fused reduce-scatter / all-reduce epilogue (`epi_reduce_mode`)

## Summary

Adds an opt-in tensor-parallel path to the SM100 GEMM that fuses the TP reduce-scatter (or
all-reduce) into the GEMM epilogue. A dedicated reducer warp group reduces each rank's output
slab in place with multimem (NVLS) load-reduce / store, offloading the cross-rank reduction to
the NVSwitch. The reduction overlaps the GEMM main loop and consumes tiles as they are
produced, so it replaces the separate post-GEMM collective launch, and its extra HBM round
trip, with in-epilogue communication.

Enable via both frontends (`quack.gemm.gemm` and `EpiMod.gemm`):
- `epi_reduce_mode="reduce_scatter"`: GEMM + reduce-scatter (each rank keeps its own M-slab).
- `epi_reduce_mode="all_reduce"`: GEMM + all-reduce (every rank ends with the full output).
- `epi_reduce_args=EpiReduceArguments(...)`: the comm-side tensors (multicast D, peer D views,
  tile flags, exit barrier, consumer counters); `quack.distributed.gemm_epi_reduce
  .make_epi_reduce_args` allocates and assembles them (collective symmetric alloc, reused
  across launches). Kernels compile per `(mode, world_size, rank)` — rank bakes the slab
  addressing — and both plan caches key on that triple.

Reference: [cutlass `distributed_gemm_reduce_scatter_blackwell.py`](https://github.com/NVIDIA/cutlass/blob/main/examples/python/CuTeDSL/cute/blackwell/kernel/distributed/distributed_gemm_reduce_scatter_blackwell.py).

**Notation.** *slab* = the `(M/TP, N)` tile each rank is responsible for reducing in RS/AR.

## Implementation

The feature is two optional hooks in the shared epilogue plus a small
self-contained module, wired up by `GemmSm100`.

`quack/gemm_base.py` — the shared `epilogue()` gains:

1. **`skip_epi_ops`** (constexpr): the producer path stores bare D partials through the normal
   smem+TMA path into symmetric D; EVT, C load, and aux stores are deferred to the reducer.
2. **`commit_D`** (callback): the reducer path stores D straight from registers, replacing
   `copy_D` (the two are mutually exclusive).
3. **`epilogue_split_k(..., signal_finalized_tile=...)`**: the split-K wrapper announces each
   finalized tile — fired once per output tile, on the finalizing entity, after its D partial
   TMA stores drain. Non-finalizing splits never signal.

`quack/epi_reduce.py` — the fused-comm pieces, one section each: the `EpiReduceArguments` host
contract; the reducer tile scheduler (`make_epi_reduce_tile_scheduler`: a slab-order static
persistent walk built from the CLC launch grid); the cross-launch exit barrier slot;
`signal_partial_committed` (the producer→reducer signal); and the multimem pair
`multimem_reduce_subtile` / `commit_reduced_subtile`.

`GemmSm100` (`quack/gemm_sm100.py`) binds them:

- Producer epilog warps run the epilogue via `epilogue_split_k` with `skip_epi_ops` bound and
  `signal_finalized_tile = signal_partial_committed(tile_flags_mc, tile_id, in_bounds)`.
- The reducer warp group walks the slab schedule and runs the *same* `epilogue()` with
  `load_acc_subtile = multimem_reduce_subtile` (ld_reduce across ranks into registers) and
  `commit_D = commit_reduced_subtile` (RS: `st.global` to this rank's slab; AR: `multimem_st`
  broadcast). EVT / C load / aux stores run unchanged in between, on the fully reduced tile.
- The dedicated epi-load warp stages C per epi_reduce tile along the same slab schedule.

### 1. Separate epi_reduce tile scheduler

Each rank owns the M-slab `[rank*M/TP, (rank+1)*M/TP)`. The reducer and epi-load warps walk
`(cta_M, cta_N)` tiles within that slab using their own static persistent scheduler
(`make_epi_reduce_tile_scheduler`), independent of the GEMM's CLC tile assignment. Each rank
walks `ceil(slab_M/cta_M) × ceil(N/cta_N)` tiles. See [Rationale](#rationale) for why a
separate, static scheduler.

### 2. Producer vs. consumer tiles

Producer (GEMM) and consumer (epi_reduce) tiles share the shape `(cta_M, cta_N)` but are out of
phase: producers are anchored at global row 0, consumers at the slab start. For TP=2, rank 1,
M=1032 (slab starts at row 516):

```
producer tiles (from row 0):    tile4 = rows 512-639    tile5 = rows 640-767
consumer tile  (from row 516):  tile0 = rows 516-643
```

So a consumer tile spans 1-2 producer tiles (exactly 1 when `M % (TP*cta_M) == 0`), and each
producer flag has 1-2 local consumers.

### 3. Signaling

When producers finish a tile's epilogue, they signal it by multimem-adding +1 to its
`tile_flags` entry (after the D partial TMA stores drain). A consumer spins until it has
accumulated `world_size` signals across the 1-2 producer flags its tile spans.

- **`tile_flags` and `consumer_counters` are monotonically increasing (never reset);** the wait
  condition is `flag - counter >= num_ranks`. Never resetting is required because (a) under PDL
  the next launch's producers overlap this launch's consumers, so a next-launch signal can
  arrive early; a store-0 reset would drop it and the following wait would hang forever; and
  (b) a flag has 1-2 consumers, so a per-visit `add -num_ranks` would subtract more than the
  producers added. Monotonic counters are safe: single-writer (no races), and int32 wrap is
  harmless because only differences are compared.
- A 2-CTA MMA tile is two 128-row halves; a fully OOB half stores nothing, so it never signals.

### 4. Predication

epi_reduce uses multimem, which, unlike the TMA descriptors used for C / colvec / aux, does not
handle OOB automatically, so we predicate explicitly:

- **`in_slab`**: cluster `ceil_div` can hand the tail CTA a tile coordinate past the slab; skip it.
- **`row_limit`**: the last slab tile's OOB rows are zero-filled on load-reduce (keeping the
  visit reduction exact, as VecReduce does) and skipped on store (a stray store could race the
  true owner's store).
- **`col_limit`**: same for the last N tile's OOB columns.

### 5. Split-K composition (SERIAL / PARALLEL)

Split-K lives entirely on the producer side and composes with epi_reduce through
`epilogue_split_k`: non-finalizing splits commit raw f32 partials to the split-K workspace and
never signal; the finalizing split folds the workspace into its accumulator, stores the
completed skip-epi-ops partial to symmetric D, and fires `signal_finalized_tile`. The reducer
side is split-K-blind — its slab schedule has no split dimension, its input is each rank's
already-folded full-K partial, and the epi-load warp stages C for every slab tile regardless of
`split_k`. SERIAL stays bitwise deterministic end to end; PARALLEL commits partials in arrival
order, so relaunches are not bit-identical (correctness by reference check only).

## Assertions / out of scope

- `m % num_ranks == 0` : requirement of a collective.
- `N % 16B` and **`d_major` must be n-major**: multimem reductions operate on 16 B vectors.
- **`split_k_mode=SEPARATE` is out of scope** (SERIAL/PARALLEL compose, see above): SEPARATE
  defers the full epilogue to a second reduce kernel over an f32 workspace, so the multimem
  cross-rank reduction would have to move into that second kernel too.
- **`rounding_mode` must be RN.** SR under epi_reduce is unspecified: the skip-epi-ops partial
  store has no seed wired and the reducer's final convert is round-to-nearest. Guarded on both
  frontends.
- **`gather_a` must be false.** gather_a raises the A/B load warps from 1 to 4 plus a separate
  A-index prefetch warp, effectively another warp group with register reallocation, and
  epi_reduce already claims a warp group. Combining them risks register spills, so it is out of
  scope for now.
- **No `varlen_m`**: the slab math needs a static M.
- **No acc-prepass epilogues** (`epi_needs_acc_prepass`, e.g. head_rmsnorm): EVT runs
  post-reduce in the reducer warps, so a prepass over per-rank partials would compute
  statistics of the wrong stream. Asserted after `epi_to_underlying_arguments` (the flag is
  only authoritative there).
- **Element-mode epilogue mods only** (no gated / `packed_cd`, no `swap_ab` / blockscaled /
  `ag_args` combination): guarded in `EpiMod.gemm`. Epilogue tensors (C, outputs, sinks) are
  slab-local `(m/TP, ...)` and validated against it.

### Distributed (multi-rank) tests

Run with `pytest --dist-only` (option added in `tests/conftest.py`). The flag selects mutually
exclusive sets: the default single-GPU suite deselects distributed tests and is unaffected, and
distributed tests refuse to run under pytest-xdist (each test launches its own torchrun
subprocess; xdist workers are pinned to a single GPU). pytest provides the parametrization; each
test case launches one torchrun subprocess. Both files allocate their comm tensors through
`make_epi_reduce_args`.

Tests run at world=4 (2 can hide numerical/ordering bugs; 8 occupies the full node).

Two test files:

- `test_gemm_epi_reduce.py` — GEMM+RS and GEMM+AR with the default epilogue, through the
  public `gemm()` frontend.
  - Dtype pairs (A/B in, D out): bf16/bf16, fp16/fp16, bf16/fp32 — each non-bf16 pair on its own
    hard shape (fp32 out exercises the 4-element multimem vector path).
  - Shapes: partial M/N tiles (M not a multiple of `TP*cta_M`), tiny grid (`M/TP < cta_M`),
    K residue, batch `l > 1`.
  - Split-K x epi_reduce: `split_k=2` SERIAL (baseline + hard shape) and PARALLEL, x RS/AR.
- `test_linear_gemm_epi_reduce.py` — epilogue variants over RS/AR, bf16, through `EpiMod.gemm`.
  - GEMM+Act: act ∈ {relu, gelu_tanh_approx}; extras ∈ {plain, bias, c (residual add), colvec,
    all} — decoupled ops for clean attribution, plus one combined `all` case (every op active:
    maximal smem staging and the full visit sequence).
  - GEMM+SqReduce: norm_weight always on; extras ∈ {plain, c}. The optional `mAuxOut`
    (pre-norm_weight snapshot) is currently untested.
  - GEMM+Act split-K: `split_k=2` SERIAL (full-tile and partial-tile shapes) and PARALLEL, x
    RS/AR, always with all extras so C staging stays in play. Split-K x sq is intentionally
    absent: split-K is producer-side, and the producer runs skip-epi-ops identically for
    every mod.

Each case layers distributed checks on top of the ref check, sharing one compile and one torchrun
launch:

1. **Run-to-run determinism**: relaunch on identical inputs must be bit-identical (stale
   flag/counter reuse is invisible to a single launch).
2. **Mutation loop** (`4*world` iterations): negate the additive inputs each iteration (sign
   flips are bit-exact in every dtype, so the expected output stays exactly predictable) and
   rotate a ~25 ms straggler rank, with no barrier between iterations — the kernel must own its
   own synchronization. Stale or raced values differ from the expected output bit-exactly.
3. **Flag/counter wrap**: seed `tile_flags`/`consumer_counters` just below `INT32_MAX` and
   relaunch across the wrap, verifying the difference-based wait survives int32 wraparound. The
   wrap is reachable in production: at TP=4 a flag wraps after ~5e8 launches.

The bit-exact asserts (1-3) hold for `split_k == 1` and SERIAL; PARALLEL cases run the same
loops but assert the reference check only (see §5).

Numerical correctness uses quack's convention:
`max|out - ref_fp32| < 2 * max|ref_dtype - ref_fp32| + 1e-5`, where `ref_fp32` recomputes the
same math (including the collective) in fp32 and `ref_dtype` is the same-dtype torch+NCCL
implementation. The sq partials are asserted per tile slot — a row-level fold accumulates
rounding bias and dilutes localized defects.

### Single-process pytest (CI guardrails, no multi-GPU)
- Non-regression for `epi_reduce_mode=None`: the default single-GPU GEMM path is unchanged
  (cubin-identical; existing GEMM tests pass).

## Rationale

**Why a separate, static scheduler for epi_reduce instead of sharing the GEMM's?**

**(a) Why not share one scheduler across the GEMM and epi_reduce (as the reference does)?** The
reference cutlass kernel reuses the GEMM's `StaticPersistentTileScheduler` in the epilogue-reduce
warp groups, remapping each tile id to a smaller `(cta_M/TP, cta_N)` sub-tile inside the rank's
slab. That remapping has three drawbacks:

- It implicitly requires M to be a multiple of `tile_shape_m * TP` (e.g. 256×8 = 2048), which is
  too restrictive for our applications.
- The remapped sub-tile's M shrinks with TP (64 / 32 / 16 at TP=2 / 4 / 8), making register
  usage harder to bound. A static scheduler always walks full `(cta_M, cta_N)` tiles, and we
  reduce internally in `(32, cta_N)` epi_reduce tiles.
- It makes the tile walk harder to reason about.

**(b) Why static, rather than quack's CLC scheduler?** Coordinating two warp groups to claim the
same tile under CLC would need extra SMEM pipelining across quack's STATIC / DYNAMIC / CLC paths;
we avoid that complexity for now.

## Benchmark
Benchmarking script is provided in `benchmarks/benchmark_gemm_epi_reduce.py` (`--mode
reduce_scatter | all_reduce`).

B300, bf16, TP=8 (GPUs 0–7), M=8192, K-sharded (k_local = K/8), tile 256×256, cluster (2,1).
Each cell **TFLOP/s (per-rank ms)**, TFLOP/s = 2·M·N·(K/TP)/t; do_bench_all warmup=5 / rep=30,
rank-synced. Floors are ceilings, not methods: **quack GEMM** = the K-sharded GEMM alone
(1 GPU, m=8192, k=K/8, same tile/cluster); **multimem RS/AR** = the standalone multimem
collective at (m, N), K-independent (one time per N, converted into each row's TFLOP/s
metric). **bound** = whichever floor is slower.

### GEMM + reduce-scatter

| proj | model | N | K | GEMM+RS | cuBLAS+NCCL | quack GEMM | multimem RS | bound |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | --- |
| output | 8B | 4096 | 4096 | 269 (0.128) | 204 (0.168) | 1111 (0.031) | 282 (0.122) | comm |
| output | 70B | 8192 | 8192 | 604 (0.228) | 470 (0.292) | 1599 (0.086) | 637 (0.216) | comm |
| output | 405B | 16384 | 16384 | 1296 (0.424) | 785 (0.701) | 1697 (0.324) | 1374 (0.400) | comm |
| down | 8B | 4096 | 14336 | 845 (0.142) | 564 (0.213) | 1601 (0.075) | 986 (0.122) | comm |
| down | 70B | 8192 | 28672 | 1540 (0.312) | 1019 (0.472) | 1695 (0.284) | 2229 (0.216) | math |
| down | 405B | 16384 | 53248 | 1687 (1.059) | 1283 (1.393) | 1658 (1.078) | 4467 (0.400) | math |

1.3–1.7× over cuBLAS+NCCL on every shape — at N=4096 the fused kernel undercuts even
standalone NCCL RS (0.128 vs 0.142 ms); the output shapes are comm-bound (the fused kernel
hides the GEMM behind the collective), down-70B/405B are math-bound. Fused act-epilogue
variants (relu/silu/gelu) cost ≤ 3 µs on every shape.

### GEMM + all-reduce

| proj | model | N | K | GEMM+AR | cuBLAS+NCCL | quack GEMM | multimem AR | bound |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | --- |
| output | 8B | 4096 | 4096 | 200 (0.172) | 114 (0.300) | 1131 (0.030) | 213 (0.161) | comm |
| output | 70B | 8192 | 8192 | 437 (0.314) | 295 (0.465) | 1648 (0.083) | 461 (0.298) | comm |
| output | 405B | 16384 | 16384 | 912 (0.603) | 545 (1.009) | 1781 (0.309) | 948 (0.580) | comm |
| down | 8B | 4096 | 14336 | 669 (0.180) | 352 (0.342) | 1656 (0.073) | 747 (0.161) | comm |
| down | 70B | 8192 | 28672 | 1420 (0.339) | 738 (0.652) | 1793 (0.268) | 1614 (0.298) | comm |
| down | 405B | 16384 | 53248 | 1651 (1.082) | 1061 (1.684) | 1785 (1.001) | 3081 (0.580) | math |

1.5–1.9× over cuBLAS+NCCL, within 4–14% of the multimem AR floor on every comm-bound shape;
down-405B is math-bound. Output is bitwise-identical to the NCCL baseline (NVLS all_reduce
f32-accumulates like the kernel's two-shot).


quack/gemm_tp_base.py
```diff
@@ -0,0 +1,545 @@
...
+    gdx, gdy, _ = cute.arch.grid_dim()
+    cl_m, cl_n = cluster_shape_mn
+    cluster_id = bidx // cl_m + (gdx // cl_m) * (
+        bidy // cl_n + (gdy // cl_n) * bidz
+    )
+    return cluster_id, bidx % cl_m, bidy % cl_n
+
+
+@cute.jit
+def static_scheduler_coord_to_slot(
```
thakkarV: prefer using layouts for this rather than manually mapping. Also, what does "slot" mean?

quack/gemm_tp_base.py
```diff
@@ -0,0 +1,545 @@
...
+    Convert this CTA's launch position to the static scheduler coordinate:
+    (linear persistent cluster id, CTA m in cluster, CTA n in cluster).
+    """
+    bidx, bidy, bidz = cute.arch.block_idx()
+    gdx, gdy, _ = cute.arch.grid_dim()
+    cl_m, cl_n = cluster_shape_mn
+    cluster_id = bidx // cl_m + (gdx // cl_m) * (
+        bidy // cl_n + (gdy // cl_n) * bidz
+    )
+    return cluster_id, bidx % cl_m, bidy % cl_n
```
thakkarV: prefer layouts for mapping and use fast divmod

quack/gemm_tp_base.py
```diff
@@ -0,0 +1,545 @@
...
+@cute.jit
+def epi_reduce_exit_slot(params: EpiReduceSchedulerParams) -> Int32:
+    # Keep the block_idx-derived coords inside one jit: returning the tuple to the
+    # kernel and re-consuming it in a second jit mis-materializes the slot -> OOB write.
+    cluster_shape_mn = params.tile_sched_params.cluster_shape_mn
+    cluster_id, cta_m, cta_n = clc_block_to_static_scheduler_coord(cluster_shape_mn)
+    return static_scheduler_coord_to_slot(cluster_id, cta_m, cta_n, cluster_shape_mn)
+
+
+def _multimem_ld_reduce_128b(dtype):
```
thakkarV: this should go into some central asm wrapper / util file.

quack/gemm_tp_base.py
```diff
@@ -0,0 +1,545 @@
...
+    EpilogueParams = ParamsBase
+
+    # Staging tiles under use_epi_reduce (read by quack.epi_utils.setup_epi_tensor):
+    #   D partials -> epi_tile (epilogue warps, unchanged)
+    #   C / EpiOp aux -> epi_reduce_tile (epi_reduce warps)
+    use_epi_reduce = None
+    epi_reduce_tile = None
+
+    @mlir_namedtuple
+    class EpiReduceArguments(NamedTuple):
```
thakkarV: can all of this epi reduction machinery be unified with the split K reduction machinery if possible?

thakkarV:
recently Tri added `distributed` dir, considering locating the dist GEMM specified files under there and reusing that machinery/design if possible

quack/gemm_tp_base.py
Nicyzk: I've removed `GemmTpTmaBase` completely:  `epilogue_skip_evt` folded into `epilogue` with a `skip_evt` option, which I think SEPARATE can use too. `epilogue_reduce` also folded into `epilogue` using `commit_D` with multimem callbacks. Now, `epi_reduce` can be composed with split-k SERIAL/PARALLEL (not SEPARATE, would require `epi_reduce` in 2nd kernel).

If you mean, unifying epi_reduce to reuse the split-K interface (i.e. tile_flags_mc + consumer_counters -> semaphore, mD_mc -> workspace). I tried but found it difficult for a few reasons:
(1) split-k uses fp32 + rounded buffer, D is fp16 
(2) split-k allocates buffers per-call, D is allocated outside kernel with nvshmem
(3) split-k (SERIAL/PARALLEL) has 1 consumer per-tile and it's one of the producers, epi_reduce can have 2 consumers per output tile, and consumer is on different rank/cta, which motivates using tile_flags_mc & consumer_counters over turnstile semaphore.

Does that make sense? Happy to iterate

quack/gemm_tp_base.py
Nicyzk: This refers to a per-CTA slot in the exit sync barrier, i've removed this function, and added a comment, hopefully clearer

quack/gemm_tp_base.py
Nicyzk: fixed

quack/gemm_tp_base.py
Nicyzk: moved to quack/dist_utils.py

quack/gemm_tp_base.py
Nicyzk: I'm planning to rethink the design again so no need to look now, will update when ready

Nicyzk: i've updated the design `epilogue_split_rank` uses a split-k protocol layered over `epilogue_split_k`, which keeps split-k function pure. i didn't merge literally as the mechanisms differ (sync primitives, buffer layout etc). lmk your thoughts

PR #177, opened 2026-07-20
[Dist GEMM] Blockscaled AllGather+GEMM

thakkarV:
- Adds `BlockScaledAllGatherRunner` (`quack/distributed/block_scaled_allgather_gemm.py`): an `AllGatherRunner` subclass that gathers blockscaled scale factors under the **same arrival flags** as the quantized values, so MXFP8/MXFP4 AG+GEMM needs no separate (exposed) SFA all-gather.
- **Zero kernel changes** — the existing AG gate (`ag_wait_m_tile`) already precedes all of a tile's TMA loads, SFA included. This PR is transport-only.
- Interface stays the parent's, made type-symmetric: `gather(X)` yields `(X_gathered, ag_args)`. A plain `Tensor` X runs the parent byte-only gather verbatim; a `BlockScaledOperand` X engages the SFA lane and yields a gathered `BlockScaledOperand`.
- Format-generic on the A side via `BlockScaledFormat` (byte geometry, scale dtype, sf vec size); B never touches the runner, so mixed AxB format pairs work exactly as in the dense GEMM.

## Performance

GB300 NVL72 (1200W), M=N=16384, K=8192 (strong scaling: fixed global shape), CUDA-graph replay, µs/iter (attained %SOL in parentheses, referenced to each method's MMA dtype peak; TP8/16 span 2/4 NVLink-connected nodes):

| TP | quack mx staged (this PR) | torch mx   | asyncTP fp8row | asyncTP bf16 | quack bf16 AG | SOL (fp8 / bf16) |
| -- | ------------------------- | ---------- | -------------- | ------------ | ------------- | ---------------- |
| 2  | **1666 (53%)**            | 1773 (50%) | 2175 (40%)     | 3288 (53%)   | 3127 (56%)    | 880 / 1759       |
| 4  | **1700 (52%)**            | 1804 (49%) | 2186 (40%)     | 3262 (54%)   | 3205 (55%)    | 880 / 1759       |
| 8  | **1661 (53%)**            | 1886 (47%) | 2138 (41%)     | 3130 (56%)   | 3125 (56%)    | 880 / 1759       |
| 16 | **1720 (51%)**            | 1927 (46%) | 2216 (40%)     | 3219 (55%)   | 3191 (55%)    | 880 / 1759       |

- Fastest at 15 of 16 points across TP 2–16 x four shapes (M 8–16k, N 4–16k, K 8k): 1.06–1.63x over torch-native mxfp8 (`all_gather` + `torch._scaled_mm`), 1.14–1.31x over async-TP's rowwise fp8, ~1.9x over bf16 AG+GEMM in absolute wall clock.
- Overhead over the dense mxfp8 roof is 1.09–1.20x and stays **flat from TP2 through TP16**, while torch-native mx degrades with rank count (its exposed NCCL gathers grow).
- async-TP has no mxfp8 path today: `fused_all_gather_scaled_matmul` validates the A scale against the shard rowcount (rowwise-only contract) and rejects blocked e8m0 scales so this is not quite an apples to apples comparision FWIW.

## Implementation details

- **Layout**: the hardware-fixed blocked SF layout `(rm=M/128, rk, 32, 4, 4)` is contiguous with the 128-row M-block dim outermost, so for `shard_m % 128 == 0` each rank's packed shard SFA is exactly a contiguous byte slice of the full packed tensor at `rank * sfa_shard_bytes`. Shards concatenate — one CE memcpy per peer assembles the gathered SFA, no repacking.
- **Ordering**: SFA peer sends are enqueued on the parent's `push_stream` *during* the `gather()` body, i.e. before the parent's exit enqueues the `[A send, flag write]` pairs. Same-stream copy engine order places every peer's SFA memcpy before that peer's flag write, so **one flag publishes both payloads**.
- **Reuse & capture safety are inherited**: the SFA buffer rotates on the same 2-buffer parity, and the parent's `ev_reuse` (all ranks' GEMM done, barrier-aggregated) signals for SFA just like it does for A. The staging event is recorded and waited within the same call and the sends ride the parent-joined push stream, so whole calls stay outer-CUDA-graph capturable with no new special cases.
- **Quantization is the caller's job**: 1x32 K-blocks are row-independent, so per-shard quantization of an M-sharded tensor is bit-identical to quantizing the gathered tensor. Quantize locally, then gather.
- **Operand contract**: `BlockScaledOperand` construction enforces qdata-dtype-vs-format and the blocked scale atom; the gathered operand carries the format's dtypes, so `validate_blockscaled_sf` still catches a runner/GEMM format mismatch at the call site.
- **Tests**: gathered bytes byte-identical to NCCL references, gated GEMM bitwise-equal to the dense GEMM, the blocked-SF shard-concat property, delayed-transport gate stress, plain-tensor fall-through, and an odd-call-count outer capture with mutated-input tests plus the full 3x3 mixed AB type format matrix.


tridao (approved):

quack/distributed/block_scaled_allgather_gemm.py
tridao, line 1: should we just put this in the same file as allgather_gemm.py?

thakkarV: yes I can fold it in.

PR #164, opened 2026-06-30
Avoid pre-electing bulk G2S copies

dongxiao92:
Follow-up to https://github.com/Dao-AILab/flash-attention/issues/2677(Finding 2) and https://github.com/Dao-AILab/flash-attention/pull/2689.

## Summary

`cute.copy` now emits the required single-lane election internally for async bulk copy atoms that lower through the `cute.copy` path. This PR removes the redundant outer `cute.arch.elect_one()` around Quack's G2S `cute.copy(CopyBulkG2SOp)` helper in `cpasync_bulk_get_copy_fn`.

Keeping both elections nests `elect_one()` around the bulk G2S copy. That can produce functionality issues once the CUTLASS DSL lowering also emits the election. Direct async bulk instructions that bypass `cute.copy`, such as the inline-PTX `cp.reduce.async.bulk` S2G path, still need their explicit election and are intentionally left unchanged here.

## SM90 Backward Finding

We hit failing FlashAttention SM90 backward varlen kernel loads LSE/dPsum through
Quack's `cpasync_bulk_get_copy_fn(gmem, smem)`. With the old Quack wrapper, that path pre-elects before calling `cute.copy(CopyBulkG2SOp)`. On a CUTLASS DSL build where `cute.copy` already emits the bulk-copy election, this creates nested election around the G2S TMA copy and reproduces the reported Hopper
failure as `cudaErrorIllegalInstruction (error code: 715)` during backward.

Removing the Quack-side wrapper lets the compiler-owned election issue the
async bulk copy from the expected context and fixes the SM90 backward case.

## Validation

Focused FlashAttention test:

```bash
flash-attention/tests/cute/test_flash_attn.py::test_flash_attn_varlen_output[True-True-True-True-random-1024-1023-192-False-True-0-0.0-False-False-False-mha-dtype0]
```

Environment:

- GPU: H200 / `sm_90a`
- FlashAttention: `6f644f9eb89c0572887986187cb0efca285a50f5`
- FlashInfer: `2099bac460994d2e6532935f1ad760a79a153f52`
- CUTLASS DSL package: `nvidia-cutlass-dsl=4.6.0a0+20260626210442.fb699d4`

Results:

- Unpatched reported Quack SHA `0735c7d47572bff49b334d5edf35731dffbd2465`:
  hit `cudaErrorIllegalInstruction (error code: 715)`.
- This PR / fix commit `af6bc485c1abbf56686dd25cf21d8adc5703370a`: same
[...]

PR #181, opened 2026-07-23
Revert electone clean up in PR-#164

dongxiao92:
Reverts #164 (commit abf5b4b).

PR #164 removed the explicit `cute.arch.elect_one()` election around the G2S `cute.copy(CopyBulkG2SOp)` calls in `cpasync_bulk_get_copy_fn`, assuming the CUTLASS DSL `cute.copy` lowering emits the required single-lane election internally for async bulk copy atoms.

After further discussion, we're reverting that change and restoring the explicit election, so these G2S bulk-copy paths stay correct on DSL builds where `cute.copy` does not emit the election itself. This restores the `with cute.arch.elect_one():` wrappers around both `copy_bulk` and `copy_bulk_single_stage`. The S2G `cp.reduce.async.bulk` path was never changed and is unaffected.

Sorry for the late changes.


tridao:
This would require a new cutlass-dsl version I assume? So we should wait for that before merging?

dongxiao92:
> This would require a new cutlass-dsl version I assume? So we should wait for that before merging?

Yes. It's planned to be 4.6.2. But the wheel is not ready yet. Does it sound good to cover the bump in the same PR once the wheel goes public?

dongxiao92:
@tridao CUTLASS DSL 4.6.2 is now live. I’ve updated this PR to pin both the base and cu13 dependencies to 4.6.2 and refreshed the branch onto current main. Could you please review it? Thanks!

dongxiao92:
Try to provide more comments for the bumping up to 4.7 for referece:
1. This Quack PR #181 lands.
2. A new Quack package containing that fix and allowing CUTLASS DSL 4.7.0.
3. FA4 updates its Quack dependency to require that new package.
5. FA4 separately updates its CUTLASS DSL dependency to 4.7.0.

tridao (approved):

Issue #182, opened 2026-07-31
[Blackwell] Is CUDA 12.9 + r575 driver still a supported combination? torch ≥ 2.12's Triton compiles sm_100 kernels with a CUDA 13.1 ptxas that r575 cannot load


JiaxuanBai:
On Blackwell GPUs (B200/GB200/B300, sm ≥ 100) with CUDA 12.9-era r575 drivers, every test that exercises `torch.compile` fails when the Inductor-compiled Triton kernel is loaded. (returns `CUDA_ERROR_INVALID_IMAGE (200)`)

The failure is not in quack's own kernels; it follows from quack's requirement for torch. Current quack requires torch ≥ 2.13 (`quack.blockscaled` calls `torch.compile(..., recompile_limit=64)` at import time, and the `recompile_limit` kwarg only exists since torch 2.13 ). torch 2.13 in turn pins triton 3.7.1, and since triton 3.7.x the bundled Blackwell assembler ([ptxas-blackwell](https://github.com/triton-lang/triton/blob/release/3.7.x/cmake/nvidia-toolchain-version.json)) is CUDA 13.1: every cubin Triton produces for `sm ≥ 100` is a CUDA 13.1 binary, which r575 (CUDA 12.9) drivers cannot load.

A large part of quack's test suite uses `torch.compile`, so in practice the suite no longer runs on Blackwell + r575. E.g., [test_cross_entropy.py](https://github.com/Dao-AILab/quack/blob/2307ab809039333b193ee17250f9475ca36ad511/tests/test_cross_entropy.py).

Could you confirm whether Blackwell + CUDA 12.9/r575 drivers are still an intended supported combination? Thanks!

**Environment**

- GB200 (aarch64) and B200 (x86_64), driver 575.xx (r575 / CUDA 12.9 branch)

tridao:
Sure we can support that (since the quack kernels still work?). I guess we'll just need to change the torch compile kwags?

JiaxuanBai:
Yes, that'd be great. Quack's own kernels are unaffected — only the Inductor/Triton-compiled paths fail. Making the torch-2.13-only compile kwargs version-compatible (`recompile_limit=` in `quack/blockscaled/quantize.py` and `quack/blockscaled/utils.py`) would restore running on torch 2.11.
Happy to send a PR for the kwarg guard if useful.

JiaxuanBai:
[...]

PR #183, opened 2026-08-03
[Blockscaled] Tolerate torch < 2.13 lacking the recompile_limit compile kwarg

JiaxuanBai:
https://github.com/Dao-AILab/quack/issues/182

torch.compile() only grew the per-wrapper recompile_limit kwarg in torch 2.13; on older torch the module-level torch.compile(...) wrappers in blockscaled/quantize.py and blockscaled/utils.py raised 'TypeError: compile() got an unexpected keyword argument recompile_limit' at import time, breaking even test collection.

Drop the kwarg when the running torch does not support it (dynamo then silently falls back to eager past 8 distinct shapes — slow but correct) and reuse _COMPILE_KW for the pack-scale wrapper so both sites share the guard.

### DIFF
--- a/quack/blockscaled/quantize.py
+++ b/quack/blockscaled/quantize.py
@@ -16,6 +16,8 @@ bf16 loss parity — FLOOR clips block maxima in (448, 512)·2^e, which hurts
 especially on gradient tensors. The FP4 quantizers remain FLOOR-only.
 """
 
+import inspect
+
 import torch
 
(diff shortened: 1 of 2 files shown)


JiaxuanBai:
Similar changes in https://github.com/Dao-AILab/quack/commit/e19f100cf1c8c9a4ec4c2a87d00f3275f7ecf087.
Close this PR.

PR #186, opened 2026-08-12
Support CUDA 12.9 stream capture info

JiaxuanBai:
## Summary

- accept both the five-value CUDA 12.x and six-value CUDA 13.x `cudaStreamGetCaptureInfo` payloads
- preserve capture dependency restoration for the all-gather CUDA Graph path
- add focused tests for both binding signatures and unexpected payloads

## Root cause

`AllGatherRunner.gather()` assumed the CUDA 13.x binding shape, including `edgeData_out`. `cuda-bindings==12.9.7` omits that field, so the six-value unpack raised `ValueError` and CUDA Graph teardown subsequently reported `cudaErrorStreamCaptureUnjoined`.

## Verification

- `py_compile` passes for the modified source and test files
- focused compatibility assertions pass for five-value, six-value, and invalid payloads
- full pytest execution was not available in the local environment

See https://github.com/Dao-AILab/quack/issues/187

### DIFF
--- a/quack/distributed/all_gather_gemm.py
+++ b/quack/distributed/all_gather_gemm.py
@@ -272,6 +272,20 @@ def _check(err_ret):
     return ret[0] if len(ret) == 1 else tuple(ret)
 
 
+def _capture_dependencies(capture_info):
+    """Extract capture dependencies from CUDA 12.x and 13.x bindings."""
+    if len(capture_info) == 5:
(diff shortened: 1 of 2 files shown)


JiaxuanBai:
Hi @tridao @Nicyzk ! Could you help review this PR? Thanks!

PR #192, opened 2026-08-26
fix: compat with cuda-bindings 12.9 and CUTLASS DSL signature change (#187, #188)

XFDG:
# Fix quack #187 + #188: cuda-bindings 12.9 & CUTLASS DSL compat

## PR URL
https://github.com/XFDG/quack/pull/new/fix-issue-187-188

## Branch
`fix-issue-187-188` on `XFDG/quack` (fork of `Dao-AILab/quack`)

---

## PR Title
`fix: compat with cuda-bindings 12.9 and CUTLASS DSL signature change`

### Summary

Fix two independent compatibility issues that block CUDA Graph capture on systems
with `cuda-bindings >= 12.9` or updated `cutlass-dsl` wheels:

1. **#187** — `AllGatherRunner.gather()` unpacked `cudaStreamGetCaptureInfo` as a
   6-tuple, but `cuda-bindings >= 12.9` returns 5 values (the `edgeData` field was
   removed), causing `ValueError: not enough values to unpack (expected 6, got 5)`.

2. **#188** — `mbarrier_arrive_release_cluster()` passed `remote_ptr` as a
   positional argument to `prims.mbarrier_arrive()`.  Current `cutlass-dsl` wheels
   require the barrier pointer as a keyword argument (`mbar_ptr=`), causing
   `TypeError: mbarrier_arrive() takes 1 positional argument but 2 were given`.

Both issues surface during CUDA Graph capture (decode path in inference engines
such as SGLang / vLLM).

---

### Root cause

| Issue | File | Line | Old behaviour | New behaviour |
|-------|------|------|---------------|---------------|
| #187 | `quack/distributed/all_gather_gemm.py` | ~801 | `_, _, _, deps, _, num_deps = _check(...)` | star-unpack + `rest[-1]` |
| #188 | `quack/pipeline.py` | ~60 | `prims.mbarrier_arrive(remote_ptr, ...)` | `prims.mbarrier_arrive(mbar_ptr=remote_ptr, ...)` |

### Changes

```diff
# quack/distributed/all_gather_gemm.py
-                    _, _, _, deps, _, num_deps = _check(
-                        runtime.cudaStreamGetCaptureInfo(compute_stream.cuda_stream)
-                    )
+                    # cuda-bindings >= 12.9 returns 5 values (edgeData removed);
+                    # older versions return 6. Star-unpack so both work.
+                    _, _, _, deps, *rest = _check(
+                        runtime.cudaStreamGetCaptureInfo(compute_stream.cuda_stream)
[...]

PR #194, opened 2026-08-28
Skip SMEM for small N in softmax forward (#62)

### LINKED ISSUE #62: Skip SMEM for small $N$ in softmax kernel
Hi, I noticed that the softmax kernel always routes data through shared memory (global -> shared -> registers), even for small row sizes where data fits entirely in registers. This adds a bit of overhead

I've experimented with a simple change for $N \le 8192$, where we instead load directly from global memory into registers, using the synchronous `CopyUniversalOp()` instead of the async shared memory path that uses `cpasync.CopyG2Op()`. To the best of my knowledge, `CopyUniversalOp()` derives from `CopyOp()` which compiles down to `ld.global` PTX instructions (and allows one to load from global memory directly to registers). 

Here are my benchmark results (H200, $M = 8192$, bf16, mean of $n=10$ runs)
| $N$ | Before | After | Delta | 
| - | - | - | - | 
| 1024 | ~2350 GB/s | ~2434 GB/s | + 50-100 GB/s |
| 2048 | ~3003 GB/s | ~3042 GB/s | Minimal |
| 4096 |  ~3419 GB/s | ~3578 GB/s | +50-100 GB/s | 
| 8192 | ~3685 GB/s | ~3800 GB/s | +100-150 GB/s | 

Would you be open to a PR for this? 

Happy to provide more benchmark data / adjust the approach based on feedback.

XFDG:
Fixes #62.

For `N <= 8192` (single-cluster case) the whole row already fits in registers and there is no other work to overlap the load with, so staging `global -> SMEM -> register` through `cp.async` (with its `commit_group` / `wait_group` round-trip) is overhead rather than latency-hiding. This PR gates a `direct_load` compile-time path in `Softmax.kernel` that loads `global -> register` directly via the synchronous `CopyUniversalOp` (`ld.global`), skipping both the SMEM allocation and the `cp.async` round-trip.

The SMEM `cp.async` path is unchanged for `N > 8192` and for the clustered (`cluster_n > 1`) case, where overlapping the load with the multi-CTA reduction still pays off. The `cluster_n == 1` clause in the gate is a safety guard — `_set_cluster_n` already returns `cluster_n == 1` for every dtype/arch at `N <= 8192`.

**Correctness:** in the direct path `cp.async`'s automatic OOB-zero-fill is not available, so OOB register lanes are pre-filled with `-inf` (via `fill_oob`) before the predicated load. The predicated copy then writes only the in-bounds lanes, leaving `-inf` in the OOB lanes so the row-max reduction stays correct on non-even-`N` rows. OOB *rows* remain harmless because the output store is already guarded by `if tXcX[0][0] < shape[0]`. The direct path never regresses any dtype, so gating on `N <= 8192` for all dtypes is safe and keeps the kernel simpler than a dtype-conditional gate.

## Changes

- `quack/softmax.py`: add the `direct_load` const_expr gate and branch the load path (direct `ld.global` vs the unchanged SMEM `cp.async` path).
- `tests/test_softmax.py`: add `test_softmax_fwd_direct_load` covering the `N=8192` (direct) / `N=8448` (SMEM) boundary and the non-even-`N` predicated OOB fill (`N=760, 3000`) across bf16/fp16/fp32.

## Testing

Verified on a dedicated (idle) H200:

- Full `tests/test_softmax.py` passes **128/128** — fwd + bwd numerics, extreme values, numerical stability, both load paths, empty batch.
- `ruff check` / `ruff format` clean.

## Performance

Dedicated H200, strict interleaved A/B (GPU-side burner so sub-100 µs kernels don't time the host enqueue gap; before/after interleaved at single-launch granularity; CPU pinned; 11–15 rounds; medians). Noise-canary rows take the SMEM path in **both** variants, so their delta must be ≈0 — it is (±0.4%), which validates the signal.

| dtype | M | N | before GB/s | after GB/s | Δ |
|-------|------|------|-------------|------------|-------|
| **fp32** | 2048 | 2048 | 4093 | 4710 | **+15.1%** |
| **fp32** | 2048 | 4096 | 3095 | 3487 | **+12.7%** |
| **fp32** | 1024 | 4096 | 3926 | 4386 | **+11.7%** |
| **fp32** | 1024 | 8192 | 3118 | 3466 | **+11.1%** |
| **fp32** | 2048 | 8192 | 3567 | 3751 | +5.2% |
| fp32 (canary, SMEM both) | 2048 | 16384 | 3880 | 3880 | −0.0% |
| bf16 / fp16 | 8192 | 1024–8192 | — | — | within ±0.2% (no change) |

The win is **fp32-specific (+5–15%)**: fp32's copy vector size is 4 (vs 8 for 16-bit), so its `cp.async` staging is relatively more expensive and skipping the SMEM round-trip pays off. bf16/fp16 land within the ±0.2% noise floor — their SMEM path is already bandwidth-efficient, so the direct path neither helps nor hurts.


XFDG:
**Note on the red `test / lint`:** it fails at the `Run actions/checkout@v6` step — `Install ruff` / `Ruff check` / `Ruff format` are all *skipped*, so the code is never actually linted. The reusable `_test.yml` is invoked from a `pull_request_target` trigger in `ci-pr.yml`, and `actions/checkout@v6` refuses to check out fork PR code in that trusted context ("pwn request" protection) unless opted in. This is a CI-config issue independent of the diff and cannot be fixed from the fork side (editing the fork's workflow has no effect — `pull_request_target` always uses the base repo's workflow file). It resolves once the checkout step in `.github/workflows/_test.yml` sets `allow-unsafe-pr-checkout: true` (or the PR runs under the `pull_request` trigger). Same situation as #192.

For what it's worth, I verified the diff locally against the CI-pinned `ruff==0.11.13`: `ruff check quack/ tests/` and `ruff format --check quack/ tests/` both pass. I also ran the full `tests/test_softmax.py` on an H200 — **128/128 pass** (fwd + bwd numerics, both load paths, empty batch). Happy to adjust anything on the code side.

PR #203, opened 2026-09-11
Forward CUTLASS converter constexpr keyword

jcui2:
CUTLASS 4.7 recursively calls `_convert_single_arg(..., is_constexpr=...)` for dataclass fields. Quack replaces that function with a four-argument wrapper, so nested argument conversion currently fails with:

```
TypeError: _patched_convert_single_arg() got an unexpected keyword argument 'is_constexpr'
```

Accept and forward the keyword to CUTLASS for its explicit constexpr path while preserving Quack's annotated-`Constexpr` and NamedTuple handling. Includes a regression test for the converter ABI.

Validated with Ruff and the focused pytest.


jcui2:
Closing at requester direction; this compatibility change will be handled in the consuming monorepo instead.

PR #210, opened 2026-09-21
Fix unbatched blocked scale unpacking

adenzhou1350:
## Summary

- accept the unbatched `(rm, rk, 32, 4, 4)` scale layout already allowed by `check_blocked_scale_atom`
- normalize it to the existing batched representation before unswizzling
- add a round-trip regression test for both rank-5 and rank-6 inputs

## Tests

- `pytest tests/test_blockscaled_quantize.py -q` (13 passed)
- `pre-commit run --files quack/blockscaled/quantize.py tests/test_blockscaled_quantize.py`


adenzhou1350:
I checked the lint logs for this PR and #209. Both jobs stop in actions/checkout before lint starts: checkout refuses fork code in the pull_request_target workflow with a privileged token. This is a workflow trust-boundary block, not a lint finding in either patch. Could a maintainer run the checks through the repository's safe fork-PR path or share the intended review/CI procedure? I will not request an unsafe checkout override.

